PR #6, opened 2017-03-01
Index

jeammimi:
Add an index property to Scatter to add an extra dimension such as the time
to x,y,z array and to be able to use it in an embed version.


js/src/volume.js
```diff
@@ -246,6 +246,9 @@ var ScatterView = widgets.WidgetView.extend( {
                 animation_time_vz : { type: "f", value: 1. },
                 animation_time_size : { type: "f", value: 1. },
                 animation_time_color : { type: "f", value: 1. },
+                animation_time_index : { type: "f", value: 1. },
```
maartenbreddels: No need for this, this should be reflected in the shaders js/glsl/scatter_vertex.js. But we can mimic a change in 'x', so just using animation_time_x, animation_time_y, animation_time_z.

ipyvolume/volume.py
```diff
@@ -27,6 +27,7 @@ class Scatter(widgets.DOMWidget):
     vy = Array(default_value=None,allow_none=True).tag(sync=True, **array_serialization)
     vz = Array(default_value=None,allow_none=True).tag(sync=True, **array_serialization)
     selected = Array(default_value=None,allow_none=True).tag(sync=True, **array_serialization)
+    index = Integer(default_value=0).tag(sync=True)
```
maartenbreddels: Lets call it sequence_index, since there may be more indices coming, hover_index, pick_index ...

js/src/volume.js
```diff
@@ -279,6 +282,20 @@ var ScatterView = widgets.WidgetView.extend( {
         _.mapObject(this.model.changedAttributes(), function(val, key){
             console.log("changed " +key)
             this.previous_values[key] = this.model.previous(key)
+            if (key == "index"){
+              pindex = this.previous_values["index"]
```
maartenbreddels: Please use this.model.previous("index"). And if it is not defined, but x is a 2d array, we better assume it is zero. Or, maybe a better way, we don't allow sequence_index to be empty/null/None, and put it to 0 by default.

js/src/volume.js
```diff
@@ -279,6 +282,20 @@ var ScatterView = widgets.WidgetView.extend( {
...
+              this.previous_values["x"] = this.model.previous("x")[pindex]
+              this.previous_values["y"] = this.model.previous("y")[pindex]
+              this.previous_values["z"] = this.model.previous("z")[pindex]
+              var vx = this.model.previous("vx")
+              if (vx && typeof vx[0][0] != "undefined" ) {
+                this.previous_values["vx"] = this.model.previous("vx")[pindex]
+                this.previous_values["vy"] = this.model.previous("vy")[pindex]
+                this.previous_values["vz"] = this.model.previous("vz")[pindex]
+              }
+
```
maartenbreddels: To make the animation work, for every property that is a sequence (x, y, z, ...)
this.attributes_changed['x'] = ['x']
This tells that 'x' (the first one) is changed, and that after it has changed, it should remove the 'x' from the list previous_values list. Do the same for y, z, vx, etc..
(basically the previous snippet in a look for 'x', 'y', ...

js/src/volume.js
```diff
@@ -279,6 +282,20 @@ var ScatterView = widgets.WidgetView.extend( {
         _.mapObject(this.model.changedAttributes(), function(val, key){
             console.log("changed " +key)
             this.previous_values[key] = this.model.previous(key)
+            if (key == "index"){
+              pindex = this.previous_values["index"]
+              this.previous_values["x"] = this.model.previous("x")[pindex]
+              this.previous_values["y"] = this.model.previous("y")[pindex]
+              this.previous_values["z"] = this.model.previous("z")[pindex]
```
maartenbreddels: Not every x, y, z, vx, vy, vz needs to be a sequence, can also be that just x is a sequence of arrays, and the rest is not. This will happen when you do updates from the notebook like
```python
scatter.x = x2d
scatter.y = y2d
```
They will be updated in sequence.
Also, I think it should be:
```
if(_.isArray(this.model.get('x')) { // changing index effectively changes the 'x' that the shader sees
  this.previous_values["x"] = this.model.get('x')[this.model.previous('sequence_index')]
  this.attributes_changed['x'] = ['x']
} // else by changing the sequence_index you don't really change x
```

maartenbreddels:
Ok, the comments are not in the order I put them in, I hope you can follow them (the check that failed is a travis error, don't worry about it).

js/src/volume.js
jeammimi: Ok, but I think we should better do
```python
if(_.isArray(this.model.previous('x')) { // changing index effectively changes the 'x' that the shader sees
  this.previous_values["x"] = this.model.previous('x')[this.model.previous('sequence_index')]
  this.attributes_changed['x'] = ['x']
} //
because the previous index was indexing the previous x 
```

maartenbreddels: I don't think so, you can change the index, and keep x the same, this code doesn't get executed when 'x' changes, only when 'sequence_index' changes.

jeammimi: Yes you are right

js/src/volume.js
jeammimi: I agree but I don't know how to trigger the animation.
If I remove this line the changes of sequence_index are not reflected.

jeammimi:
I just updated and modified accordingly. I just kept the  animation_time_index , because if not it does not work

js/src/volume.js
maartenbreddels: Say if 'x' is a sequence, you'd set:
```javascript
this.previous_values["x"] = this.model.get('x')[this.model.previous('sequence_index')]
this.attributes_changed['x'] = ['x']
```
Now (via calling this.update, and then arriving at create_mesh)
It would (or should) detect that x has changed, and it will fill the `offsets_previous` with the this.previous_values["x"] values, and the `offsets` with the current `x[sequence_index]` values. I think that should work. Otherwise update this PR so I can see the new code.

jeammimi:
Yes you are right.

In another branch I started to play with having different color for different points
[Time + color](https://rawgit.com/jeammimi/replication_yeast/master/notebooks/report/test.html)
For  this I introduce a multi_colors component that can be 1D or 2D

maartenbreddels:
Check out (Union)[https://github.com/ipython/traitlets/blob/35b0f68f7b3652d36d06f7528748d5778c632a73/traitlets/traitlets.py#L1792], I think color should be able to be a string, a list of strings (or a sequence or a list of strings actually) (and later also support for an array of float, in combination with a colormap). Would love to see this in a PR as well! But lets first finish this one.

jeammimi:
I added the check with x, y , z on create_mesh.
 But I also merged your changes, and the new coordinate are not taken into account anymore.

jeammimi:
I added a demo notebook to try it

jeammimi:
Is it a real problem to keep animation_time_index

maartenbreddels:
Yeah, no need for it, I send you the fix for that.

maartenbreddels:
Ok, I'd like to keep this PR clean, if you drop the notebook (I have a copy, I'll make sure I'll it goes somewhere in the examples, or in the docs), and confirm that it is working, I can accept this PR.

jeammimi:
Ok its done ! And I confirm it works!

jeammimi:
Great! thank you!

maartenbreddels:
Thank you, check this out:
https://www.astro.rug.nl/~breddels/test/stream.html
Something I had in mind, but you made possible (it's a stellar stream, simulation of small galaxy getting ripped apart, you don't see the host galaxy)

jeammimi:
So cool!!!!
It is really nice to watch !
What widget did you use to have this play button?

maartenbreddels:
Thank, I didn't do the simulation btw.
See http://ipywidgets.readthedocs.io/en/latest/examples/Widget%20List.html#Play-(Animation)-widget

maartenbreddels:
It really combines well with the animation/interpolation, you can get away with a coarse timestep.

maartenbreddels:
And if you refresh it, you'll see velocities are also working, nice work :+1:

jeammimi:
It is really nice!
To make it ligther ( in terms of megabite) I multilply x ,y , ... by a scale factor, and then
I give np.array(x,dtype=np.int) to your function. If you give it the float version, it is going to keep a lot
of numbers that are not necessary.

PR #9, opened 2017-03-03
Fix width and height in figure

jeammimi:
Small fix to take into account the height and width when given to figure.
example:
```python
import ipyvolume.pylab as p3
f = p3.figure(width=1000,height=300)
```


maartenbreddels:
thx :+1:

PR #10, opened 2017-03-03
Multicolors

jeammimi:
Added the possibility that color is a string a list of string or a 2d array of string

### DIFF
--- a/js/src/volume.js
+++ b/js/src/volume.js
@@ -196,7 +196,7 @@ var ScatterView = widgets.WidgetView.extend( {
         console.log("create scatter")
 
         this.geo_diamond = new THREE.SphereGeometry(1, 2, 2)
-        this.geo_sphere = new THREE.SphereGeometry(1, 12, 12)
+        this.geo_sphere = new THREE.SphereGeometry(1, 4*12, 4*12)
         this.geo_box = new THREE.BoxGeometry(1, 1, 1)
         this.geo_cat = new THREE.Geometry(1, 1, 1)
         for(var i = 0; i < cat_data.vertices.length; i++) {
@@ -478,32 +478,58 @@ var ScatterView = widgets.WidgetView.extend( {
         // colors
         var colors = new THREE.InstancedBufferAttribute(new Float32Array( max_count * 3 ), 3, 1);
         var colors_previous = new THREE.InstancedBufferAttribute(new Float32Array( max_count * 3 ), 3, 1);
-
         function to_rgb(color) {
             color = new THREE.Color(color)
             return [color.r, color.g, color.b]
         }
-        var color = to_rgb(this.model.get("color"))
-        var color_previous = "color" in this.previous_values ? to_rgb(this.previous_values["color"]) : color;
-        if(!color_previous)
-            color_previous = color;
 
-        var color_selected = to_rgb(this.model.get("color_selected"))
-        var color_selected_previous = "color_selected" in this.previous_values ? to_rgb(this.previous_values["color_selected"]) : color_selected;
-        if(!color_selected_previous)
-            color_selected_previous = color_selected;
+        var multi_colors = this.model.get("multi_colors")
+        //console.log("multi",multi_colors)
+        if( multi_colors){
 
-	    for(var i = 0; i < max_count; i++) {
-	        var cur_color = color;
-	        if(selected.indexOf(i) != -1)
-	            cur_color = color_selected
-   	        colors.setXYZ(i, cur_color[0], cur_color[1], cur_color[2]);
+          if( (typeof multi_colors[0][0] != "undefined") && (typeof multi_colors[0] === 'string')){
+            //1D
+            console.log("1D color")
+          }
+          else {
+            //2D + index
+            console.log("2D color")
+            multi_colors = multi_colors[index]
+          }
 
-	        var cur_color_previous = color_previous;
-	        if(selected_previous.indexOf(i) != -1)
-	            cur_color_previous = color_selected_previous
-   	        colors_previous.setXYZ(i, cur_color_previous[0], cur_color_previous[1], cur_color_previous[2]);
-	    }
+          for(var i = 0; i < max_count; i++) {
+              var cur_color = to_rgb(multi_colors[i]);
+              if(selected.indexOf(i) != -1)
+                  cur_color = color_selected
+                colors.setXYZ(i, cur_color[0], cur_color[1], cur_color[2]);
+            }
+
+
+        }
+        else {
+
+          var color = to_rgb(this.model.get("color"))
+          var color_previous = "color" in this.previous_values ? to_rgb(this.previous_values["color"]) : color;
+          if(!color_previous)
+              color_previous = color;
+
+          var color_selected = to_rgb(this.model.get("color_selected"))
+          var color_selected_previous = "color_selected" in this.previous_values ? to_rgb(this.previous_values["color_selected"]) : color_selected;
+          if(!color_selected_previous)
+              color_selected_previous = color_selected;
+
+    	    for(var i = 0; i < max_count; i++) {
+    	        var cur_color = color;
+    	        if(selected.indexOf(i) != -1)
+    	            cur_color = color_selected
+       	        colors.setXYZ(i, cur_color[0], cur_color[1], cur_color[2]);
+
+    	        var cur_color_previous = color_previous;
+    	        if(selected_previous.indexOf(i) != -1)
+    	            cur_color_previous = color_selected_previous
+       	        colors_previous.setXYZ(i, cur_color_previous[0], cur_color_previous[1], cur_color_previous[2]);
+    	    }
+        }
 
         instanced_geo.addAttribute( 'color', colors );
         instanced_geo.addAttribute( 'color_previous', colors_previous );
--- a/js/src/volume.js
+++ b/js/src/volume.js
@@ -308,6 +308,15 @@ var ScatterView = widgets.WidgetView.extend( {
                 this.previous_values["vz"] = this.model.get("vz")[pindex]
                 this.attributes_changed["vz"] =["vz"]
               }
+              if (this.model.get("color") ) {
+                color = this.model.get("color")
+                if (!(typeof color == "string" || typeof color[0] == "string")){
+                //check for 2d
+                    this.previous_values["color"] = this.model.get("color")[Math.min(pindex,color.length-1)]
+                    this.attributes_changed["color"] =["color"]
+                  }
+              }
+
 
             }
 	    else if(key_animation == "geo") {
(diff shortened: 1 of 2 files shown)


js/src/volume.js
```diff
@@ -310,6 +308,7 @@ var ScatterView = widgets.WidgetView.extend( {
                 this.previous_values["vz"] = this.model.get("vz")[pindex]
                 this.attributes_changed["vz"] =["vz"]
               }
+
```
maartenbreddels: I guess here is similar code needed if color is a 2d array, if sequence_index changes, it needs to update 'color'. Make sure the animation works for this.

js/src/volume.js
```diff
@@ -485,32 +484,70 @@ var ScatterView = widgets.WidgetView.extend( {
...
+  	        if(selected.indexOf(i) != -1)
+  	            cur_color = color_selected
+
+     	        colors.setXYZ(i, cur_color[0], cur_color[1], cur_color[2]);
+
+  	        var cur_color_previous = color_previous[i];
+  	        if(selected_previous.indexOf(i) != -1)
+  	            cur_color_previous = color_selected_previous
+     	        colors_previous.setXYZ(i, cur_color_previous[0], cur_color_previous[1], cur_color_previous[2]);
+  	    }
```
maartenbreddels: This will decrease the performance for having a million points, but a single color. So I'd rather see that if it is a single color, to not create an array.

js/src/volume.js
```diff
@@ -485,32 +484,70 @@ var ScatterView = widgets.WidgetView.extend( {
...
+        }
+
+
+        var color = get_value_index_color(this.model.get("color"),index,max_count)
+
+        var color_previous = "color" in this.previous_values ? get_value_index_color(this.previous_values["color"],this.previous_values["index"],max_count) : color;
         if(!color_previous)
             color_previous = color;
+        console.log("color",color)
+        console.log("color_previous",color_previous)
```
maartenbreddels: Please remove these.

js/src/volume.js
jeammimi: Done, It didn't know that there was animation on the color!

js/src/volume.js
jeammimi: Done

maartenbreddels:
Could you fix the indenting a bit? For the rest it is working, animation etc? We could really use unittests at the js side now. But setting this up with opengl rendering won't be easy...

jeammimi:
Yes unitest would be nice, but I have no clue how to implement them.
Yes everything I could think of testing worked. 
Do you want me to add the netbook I used for testing?

slinnarsson:
Better than array of string would be numpy 2d vector of rgba values. Size (n_samples, 4). That would work out of the box with matplotlib's colormaps.

maartenbreddels:
Yes, and another option would be to have a float->color mapping on the js side (color scales). Lets see if this would be work out:
 * shape is 0 dim, and it's a string, interpret as color
 * shape is 1 dim, items are strings, seperate color for each item
 * shape is 2 dim, items are strings, sequence of the above
 * no color scale / colormap is set
  * shape is 1 dim, items are floats, it should be of length 3 -> rgb values (possibly later 4 for opacity)
  * shape is 2 dim, items are float, it should be of shape (len(x), 3) -> rgb values
  * shape is 3 dim, items are float, it should be (sequence_length, len(x), 3) -> rgb values
 * color scale / colormap is set
   * shape is 0 dim, item if float, use the colorscale to map to color
   * shape is 1 dim, it's a seperate color for each item
   * shape is 2 dim, again a sequence

So this can work out I think, no ambiguity about how color would be interpreted. The color scale is not something I want to address now. @jeammimi If you feel like implementing the rgb values by float, I'm happy to see this in this PR (hint: see [_.isNumber](http://underscorejs.org/#isNumber), otherwise @slinnarsson can give it a try?

maartenbreddels:
A different PR is fine as well, shall i accept this first?

jeammimi:
I can give it a try. For the color scale, I am not familiar with it. So I think it will be for another time.
I will implement 1D 2D 3D with shape 3.
Transparency could be nice, but I don't know where it should be specified in the javascript part.

jeammimi:
Ok it is done.
The only thing is that there is a warning that appears with traitlets when the array color is numeric.
```python
traitlets/traitlets.py:565: FutureWarning: elementwise comparison failed; returning scalar instead, but in the future will perform elementwise comparison
  silent = bool(old_value == new_value)
```

jeammimi:
So now it works for these cases:

-  shape is 0 dim, and it's a string, interpret as color
-  shape is 1 dim, items are strings, seperate color for each item
-  shape is 2 dim, items are strings, sequence of the above
- shape is 1 dim, items are floats, it should be of length 3 -> rgb values 
- shape is 2 dim, items are float, it should be of shape (len(x), 3) -> rgb values
- shape is 3 dim, items are float, it should be (sequence_length, len(x), 3) -> rgb values

js/src/volume.js
```diff
@@ -485,32 +502,103 @@ var ScatterView = widgets.WidgetView.extend( {
...
+            if ( typeof variable == "string" || typeof variable[0] == "string" || typeof variable[0][0] == "string") {
+                vartype = 1;
+            }
+
+            if (vartype == 1) {
+                if ( typeof variable == "string") {
+                //OD
+                    color =  to_rgb(variable)
+                    return color
+                }
```
maartenbreddels: I suggest refactoring this a bit, so it's easier to read/maintain.
instead of returning the color, and checking the color dim, you could do sth like this:
```javascript
color = to_rgb(variable)
return function(glyph_index) {
   return color
}

File: js/src/volume.js [hunk 5]
```diff
@@ -485,32 +502,103 @@ var ScatterView = widgets.WidgetView.extend( {
...
+                    }
+                    else if (typeof index != "undefined"  && typeof variable[0][0] == "string"){
+                        //Two D
+                        index1 = Math.min(index,variable.length - 1);
+                        to_convert = variable[index1]
+                    }
+
+                    for(var i = 0; i < max_count; i++) {
+                        tmp_color[i] = to_rgb(to_convert[i])
+                    }
```
maartenbreddels: And here:
```javascript
... convert ...
return function(glyph_index) {
   return tmp_color[glyph_index]
}
```
etc..

js/src/volume.js
```diff
@@ -485,32 +502,103 @@ var ScatterView = widgets.WidgetView.extend( {
...
-	            cur_color = color_selected
-   	        colors.setXYZ(i, cur_color[0], cur_color[1], cur_color[2]);
+  	    for(var i = 0; i < max_count; i++) {
+            if (color_d == 0){
+                var cur_color = color;
+            }
+            else{
+  	            var cur_color = color[i];
+            }
+
```
maartenbreddels: And then here you could use cur_color = color(i)

js/src/volume.js
```diff
@@ -485,32 +502,103 @@ var ScatterView = widgets.WidgetView.extend( {
...
+  	        if(selected.indexOf(i) != -1)
+  	            cur_color = color_selected
+
+     	      colors.setXYZ(i, cur_color[0], cur_color[1], cur_color[2]);
+            if (color_previous_d == 0){
+  	            var cur_color_previous = color_previous;
+            }
+            else{
+                var cur_color_previous = color_previous[i];
+            }
```
maartenbreddels: and here cur_color_previous = color_previous(i)

js/src/volume.js
```diff
@@ -310,6 +308,25 @@ var ScatterView = widgets.WidgetView.extend( {
                 this.previous_values["vz"] = this.model.get("vz")[pindex]
                 this.attributes_changed["vz"] =["vz"]
               }
+
+              if (this.model.get("color") ) {
+                  color = this.model.get("color")
+                  if ( typeof color == "string" || typeof color[0] == "string" || typeof color[0][0] == "string") {
+                      if (!(typeof color == "string" || typeof color[0] == "string")){
```
maartenbreddels: Could you make this a bit simpler, sth like:
```javascript
 if ( ... 0 or 1d check ) {
 // don't do anything, we don't have a sequence, 
}
else if ( .. 2d check ...) {
  // we have a sequence
...
}
```
I think this is more easy to read. 
Please be consistent with whitespace, no whitespace after `if(`,

js/src/volume.js
```diff
@@ -485,32 +502,103 @@ var ScatterView = widgets.WidgetView.extend( {
...
-
         function to_rgb(color) {
             color = new THREE.Color(color)
             return [color.r, color.g, color.b]
         }
-        var color = to_rgb(this.model.get("color"))
-        var color_previous = "color" in this.previous_values ? to_rgb(this.previous_values["color"]) : color;
+
+        function get_value_index_color(variable,index,max_count){
+            //Return a two D array (max_count,3)
```
maartenbreddels: I think this function should go top level, it can be reused if for instance we're gonna support lines. And instead of returning a color, I think it should return a function. (see the rest of the comments)
Could you document it, by starting with this list:
-  shape is 0 dim, and it's a string, interpret as color
-  shape is 1 dim, items are strings, seperate color for each item
-  shape is 2 dim, items are strings, sequence of the above
- shape is 1 dim, items are floats, it should be of length 3 -> rgb values 
- shape is 2 dim, items are float, it should be of shape (len(x), 3) -> rgb values
- shape is 3 dim, items are float, it should be (sequence_length, len(x), 3) -> rgb values

and then explaining how all these cases get handled. At the moment it's hard to say what is going on by quickly going over the code.

And please use this style (note the whitespace):
```javascript
function get_value_index_color(variable, index, max_count) {
```

maartenbreddels:
Lost of comments, sorry for that ;).
Color scale and transparency is another thing.. don't worry about it for the moment, at least we know that we have a list of unambigious behavious for 'scatter'.

Btw, we could think of doing unittests/test, but not on travis. Now it is possible to do a p3.savefig('test_multicolor.png') in the notebook, and compare say that image against an previous image to see we didn't mess up.
Maybe not yet for this PR, but just to let you know.

jeammimi:
Ok, I modified everything accordingly. 
The function get_value_index_color is now cleaner and easy to read I think.
The only thing I changed is that I don't create anymore the array tmp_color.
It is better for the memory, and I thing that at worse the conversion is made twice.
Here is my email jeanmichel.arbona at gmail.com

maartenbreddels:
Looking good :+1: , I'll pull this in.

jeammimi:
Great!

maartenbreddels:
I'll put something like this as an example:
https://www.astro.rug.nl/~breddels/test/wave/
I think it demonstates most of the features. Some arrays can be constant (x and z in this case), and y and color are a sequence.

maartenbreddels:
I added size as well, it can be a float or a sequence.

jeammimi:
Yes it is super nice.
Here the is also a sequence no ? You implemented it ?

maartenbreddels:
Yes, just pushed it to master.

slinnarsson:
Fantastic! Thanks for doing this. The demo looks great.

/Sten

-- 
Sten Linnarsson, PhD
Professor of Molecular Systems Biology
Karolinska Institutet 
Unit of Molecular Neurobiology 
Department of Medical Biochemistry and Biophysics 
Scheeles väg 1, 171 77 Stockholm, Sweden
+46 8 52 48 75 77 (office)
+46 70 399 32 06 (mobile)

jeammimi:
I think there is a problem with the implementation for size.
I don't know why but when size is a float, the sequence does not work.

jeammimi:
I think you should modify line 419 op volume.js with 
```python
if (this.model.get("size") && !_.isNumber(this.model.get("size")) && typeof this.model.get("size")[0][0] != "undefined" ) {
```
because if it is a number aparently it does not like being checked for an array

maartenbreddels:
Yes, i fixed that, i'll push it to master.

maartenbreddels:
fixed.

PR #17, opened 2017-03-21
Modified for the embeding to work in case of performance equal one

jeammimi:


maartenbreddels:
I wonder, can't we do sth like this in embed_html?:
```python
previous = ipyvolume.serialiize.performance
try:
   ipyvolume.serialiize.performance = 0
  ...
         for widget in widgets:
             if not all:
                 get_state(widget, state, drop_defaults=drop_defaults)
  ...
finally:
  ipyvolume.serialize.performance = previous

jeammimi:
I am not sure how it works, but this get_state get the value from the js no?

maartenbreddels:
No, it will basically call the to_json, which will give a list for performance=0, can you try that out?

jeammimi:
Are you sure, because it keep the state of the widget as it is shown.

maartenbreddels:
True, but the python side is always synced. And it will always go though a from_json path etc.

jeammimi:
ok, I ll try

jeammimi:
ok it works, I will create another pull request

PR #18, opened 2017-03-21
Embeding working with performance == 1 or 2

jeammimi:


maartenbreddels:
Nice that it's working, could you make sure that the previous value is set by putting it in a try/finally block:
```python
try:
  ....
finally:
  pass # this code will execute whatever happens
```

maartenbreddels:
👍

maartenbreddels:
merge, thanks again!

jeammimi:
You are wellcome! Now you can set performance == 1 by default :)

maartenbreddels:
I think that is true actually!
Also, the performance = 2 case will be supported for the next release of ipywidgets, https://github.com/ipython/ipywidgets/pull/1194 got merged. So then we can have sequences with non-constant length arrays. Not sure what we need to do with performance == 2, then, maybe just remove the performance stuff, and by default try to send it as binary.

PR #21, opened 2017-03-24
Added the possibility to store the data in an external file.

jeammimi:
This property is control by the attribute embed added to the Scatter class
if set to None, the data is stored in the html file
if set to "" the data is stored in a json file whose name is generated kind of randomly
if set to example.json if will be stored in example.json 
You can also control the behavior when applying embed_html
if inside is set to False, if will apply the rules previously describe.
if inside is set to True, in will put the data inside the html

### DIFF
--- a/ipyvolume/embed.py
+++ b/ipyvolume/embed.py
@@ -88,7 +88,7 @@ def add_referring_widgets(states, drop_defaults=False):
 
                                 #get_state(value, store, drop_defaults=drop_defaults)
     return found_new
-def embed_html(filename, widgets, drop_defaults=False, all=False, title="ipyvolume embed example", template=template, **kwargs):
+def embed_html(filename, widgets, drop_defaults=False, all=False, title="ipyvolume embed example", template=template, inside=False, **kwargs):
     try:
         widgets[0]
     except (IndexError, TypeError):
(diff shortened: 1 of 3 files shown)


maartenbreddels:
Hey, this is a nice idea. I wonder if we can take this a step higher. To store the whole application to a separate json file, and load that. This would possible be something that could go upstream in ipywidgets as well. 
Would that cover your use case as well? Say loading the whole widget state from an external file, or do you also need to load other files?

jeammimi:
We can store all the widget attributes, in a separate files (the one defined at tre traitlets level).
So there is not much left after that. But I don't know enough of the internal of ipywidgets to say more.

jeammimi:
What annoyed me a little with my implementation is that I could not find a way to load data from a model, and I had to use a collection. So to minimize the change on your code, this collection is an attribute of the scatterview.

maartenbreddels:
See for instance in the Notebook `Widgets->Download widget state`, that is exactly the same as the json that is put in the html. If we put this json code in an external file, and then load this (with a progress bar), and call the ipywidget library to then look at this json data and build the widgets.

@SylvainCorlay: I wouldn't be surprised if someone already did this, do you know?

jeammimi:
I see.
In my mind, I was more thinking about drawing something nice inside the scatter while loading the data :)

maartenbreddels:
Yeah, I think both cases would be interesting. For instance (thinking ahead), I'd like to see some kind of 'datastore' object on the JS side, so that multiple libraries, like bqplot, ipyvolume could tap into that data. In that case you could think of loading external data. But that is (I think) quite far into the future, so I'm not sure what is wise to do in the meantime.
The idea of loading widgets from an external json file is I think not difficult to implement, quite general, but the loading of external data may require some more thinking.

SylvainCorlay:
A widget "file format" which stores both the manager state (current json downloadable in the menu bar) and which top-level widget(s) should be displayed at a top level.

Then we can make a loader for that file format in jupyterlab.

maartenbreddels:
@SylvainCorlay what about an intermediate solution. If we add to the state schema a 'url' property. The manager could then, if this 'url' property is present, load the json using ajax, and have some kind spinner showing while loading.

SylvainCorlay:
Sounds like a good idea too.

maartenbreddels:
@jeammimi would this work for you? I cannot work on if for ~2 weeks, I could help out if you want to give it a go. But it requires work setting up a dev version of ipywidgets.

jeammimi:
Hi. I will not be able to work on it this week, but I will see next week, if it is not too hard to implement.
I am a beginner with javascript and sometimes it takes me ages to do simple stuff...

PR #32, opened 2017-07-10
Update gitignore, add top-level jupyter-widgets class

SylvainCorlay:


js/src/figure.js
```diff
@@ -60,6 +60,7 @@ var FigureView = widgets.DOMWidgetView.extend( {
         var width = this.model.get("width");
         var height = this.model.get("height");
         this.renderer = new THREE.WebGLRenderer({antialias: true});
+        this.el.classList.add("jupyter-widgets");
```
maartenbreddels: You've mentioned this before, what does this bring me?

SylvainCorlay: There is some top-level styling associated to it in core widgets. (margin etc) that will help it be aligned when used in layouts with VBox and HBox

PR #40, opened 2017-08-01
Addition of Field-of-View Hook and switched to CombinedCamera

chrisjsewell:
Hows this for a first stab?
- I've added a `camera_fov` trait and also some math so that, when changing fov, you maintain the same zoom level.
- I've switched out the Perspective Camera for the Combined Camera, to check that it still works the same in perspective mode. But haven't yet added any traits/logic for switching to/from orthographic mode

### DIFF
--- a/js/src/figure.js
+++ b/js/src/figure.js
@@ -17,6 +17,7 @@ require("./three/TrackballControls.js")
 require("./three/DeviceOrientationControls.js")
 require("./three/StereoEffect.js")
 require("./three/THREEx.FullScreen.js")
+require("./three/CombinedCamera.js")
 ndarray = require('ndarray')
 
 function is_ndarray(obj) {
@@ -70,16 +71,23 @@ var FigureView = widgets.DOMWidgetView.extend( {
         this.el_axes = document.createElement("div")
         this.el_mirror.appendChild(this.el_axes);
 
-        const VIEW_ANGLE = 45;
-        const aspect = width / height;
+        //const VIEW_ANGLE = this.model.get("camera_fov");
+        //const aspect = width / height;
         const NEAR = 0.01;
         const FAR = 10000;
-        this.camera = new THREE.PerspectiveCamera(
-            VIEW_ANGLE,
-            aspect,
+        const orthoNEAR = -500;
+        const orthoFAR = 1000;        
+        this.camera = new THREE.CombinedCamera(
+            window.innerWidth/2,
+            window.innerHeight/2,
+            this.model.get("camera_fov"),
+            //aspect,
             NEAR,
-            FAR
+            FAR,
+            orthoNEAR,
+            orthoFAR
         );
+        //this.camera.toOrthographic()
         this.camera_stereo = new THREE.StereoCamera()
         this.renderer.setSize(width, height);
 
@@ -142,7 +150,8 @@ var FigureView = widgets.DOMWidgetView.extend( {
         this.wire_box.add(make_line(-0.5, -0.5+1, -0.5, -0.5, -0.5+1, -0.5+1, this.axes_material))
         this.wire_box.add(make_line(-0.5+1, -0.5+1, -0.5, -0.5+1, -0.5+1, -0.5+1, this.axes_material))
 
-        this.camera.position.z = 2
+        // set a good intial z for any fov angle
+        this.camera.position.z = 2 * this.getTanDeg(45/2) / this.getTanDeg(this.model.get("camera_fov")/2)
 
 
         // d3 data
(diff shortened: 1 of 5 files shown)


js/src/scatter.js
```diff
@@ -20,6 +20,7 @@ var ScatterView = widgets.WidgetView.extend( {
 
         this.geo_diamond = new THREE.SphereGeometry(1, 2, 2)
         this.geo_sphere = new THREE.SphereGeometry(1, 12, 12)
+        this.geo_sphere_hres = new THREE.SphereGeometry(1, 48, 48)
```
maartenbreddels: Can you remove this, to keep this PR cleaner?

js/src/scatter.js
```diff
@@ -53,6 +54,7 @@ var ScatterView = widgets.WidgetView.extend( {
             box: this.geo_box,
             arrow: this.geo_arrow,
             sphere: this.geo_sphere,
+            sphere_hres:this.geo_sphere_hres,
```
maartenbreddels: And this as well.

js/src/figure.js
```diff
@@ -142,7 +150,8 @@ var FigureView = widgets.DOMWidgetView.extend( {
         this.wire_box.add(make_line(-0.5, -0.5+1, -0.5, -0.5, -0.5+1, -0.5+1, this.axes_material))
         this.wire_box.add(make_line(-0.5+1, -0.5+1, -0.5, -0.5+1, -0.5+1, -0.5+1, this.axes_material))
 
-        this.camera.position.z = 2
+        // set a good intial z for any fov angle
+        this.camera.position.z = 2 * this.getTanDeg(45/2) / this.getTanDeg(this.model.get("camera_fov")/2)
```
maartenbreddels: What's the idea behind this? Can you link to sth or provide a hint?

maartenbreddels:
Btw, what about the camera license, I guess it's the same as three, since it is in the threejs repo: https://github.com/mrdoob/three.js/blob/master/examples/js/cameras/CombinedCamera.js

Also, we could make pythreejs a hard dependency, so we expose the cameras object as pythreejs objects instead. What do you think about that? That would mean that you always need to install pythreejs, which would be an extra possible point of failure, which I don't like.

js/src/figure.js
chrisjsewell: Just some 'back of an envelope' trigonometry: 

fullsizerender

Since the camera is intialized at (0,0,z) (i.e. a=z) and we know that theta=45 and z=2 provides a good initial view (as is currently hard coded), hence this line.

Make sense?

chrisjsewell:
I copied it directly from there, so I assume its covered by the three.js license, since there is no sub licences in any of the directory path.

For the pythreejs integration, I think its a good long term goal. But, I guess, hold off on that until it is closer to a version 1.0

js/src/figure.js
maartenbreddels: perfect, maybe put a link to this PR.

maartenbreddels:
Yup, I agree. If you remove the high res spheres, happy to merge.

chrisjsewell:
cheers

js/src/figure.js
vidartf: Maybe put back in `const VIEW_ANGLE = 45;` and use that to avoid magic numbers?

maartenbreddels:
Thanks for you first contribution! 👍 
I'm not sure about the sphere_highres, another option would be to include support for pythreejs, but again, that would depend on pythreejs, and I see no other reason to include high res versions of other glyphs, so why not. Using normals would also make them look more smooth btw.

js/src/figure.js
vidartf: (Updated) Maybe add 45 as a constant `INITIAL_VIEW_ANGLE` (or more descriptive) to avoid magic numbers?

maartenbreddels: well, it's not really a fov anymore... it sets the initial distance. I'll think about it.

chrisjsewell:
Yeh no problem,  `sphere_hres` was obviously just a quick fix and a more comprehensive solution would be best

js/src/figure.js
chrisjsewell: Hey vidarf, I'll leave your suggestion to maarten. But, for future knowledge, what do you mean by magic numbers?

vidartf: Ah, by magic numbers I mean any number that appears in code without explanation (e.g. `3.14...`). These should typically be replaced by constant vars, or at least commented.

chrisjsewell: ah ok ta, just wanted to check it wasn't some obscure javascript issue lol

PR #43, opened 2017-08-02
works on python-3.6

stonebig:


maartenbreddels:
That is true, I developer under 3.6 actually! Thanks.

Issue #42, opened 2017-08-02
Ensure all figure traits are synced when rendering


chrisjsewell:
Sorry for flooding your issues section lol

So, for instance, setting `fig.anglex = 45` will do nothing if it is before the figure has been displayed.

Adding `this.update_current_control();` to the end of the render function fixes this issue for control traits, but is there a more general method to ensure all traits are synced?

maartenbreddels:
I think that should be the way to do it.

PR #45, opened 2017-08-06
ensure initial sync of view with figure model

chrisjsewell:
Just a quick PR to close maartenbreddels/ipyvolume#42


maartenbreddels:
looks good 👍  thanks!

PR #46, opened 2017-08-06
moved save_js code to separate function and added to embed_html

chrisjsewell:
so that `pylab.save` and `embed.embed_html` work the same, with no code duplication


maartenbreddels:
Hi Chris, looks good, and indeed this should move up to ipywidgets (i hope to get rid of most of the code in embed.py).
For github workflow, try to merge in the master in your master after it's merged, and work with different branches for the PR's. That will get rid of the long history, but for now it's ok, i'll squash and merge.

PR #49, opened 2017-08-12
added visible trait to Mesh

chrisjsewell:
Essentially acting as a binary opacity=0 or 1, until *true* opacity can be implemented

### DIFF
--- a/ipyvolume/widgets.py
+++ b/ipyvolume/widgets.py
@@ -50,8 +50,9 @@ class Mesh(widgets.DOMWidget):
 #                                     Unicode().tag(sync=True)],
 #                                     default_value="green").tag(sync=True)
 #    geo = traitlets.Unicode('diamond').tag(sync=True)
-    wire = traitlets.CBool(default_value=False).tag(sync=True)
     visible = traitlets.CBool(default_value=True).tag(sync=True)
+    visible_lines = traitlets.CBool(default_value=True).tag(sync=True)
+    visible_faces = traitlets.CBool(default_value=True).tag(sync=True)
 
 @widgets.register
 class Scatter(widgets.DOMWidget):
--- a/ipyvolume/widgets.py
+++ b/ipyvolume/widgets.py
@@ -51,6 +51,7 @@ class Mesh(widgets.DOMWidget):
 #                                     default_value="green").tag(sync=True)
 #    geo = traitlets.Unicode('diamond').tag(sync=True)
     wire = traitlets.CBool(default_value=False).tag(sync=True)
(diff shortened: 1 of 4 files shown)


js/src/mesh.js
```diff
@@ -40,7 +40,8 @@ var MeshView = widgets.WidgetView.extend( {
             fragmentShader: require('../glsl/mesh-fragment.glsl'),
             polygonOffset: true,
             polygonOffsetFactor: 1, // positive value pushes polygon further away, so wireframes will render properly (z-buffer issues)
-            polygonOffsetUnits: 1
+            polygonOffsetUnits: 1,
+            visible: this.model.get("visible")
```
maartenbreddels: This would need to be copied to all the other materials. Otherwise I would still be visible when you do texture rendering for instance.

ipyvolume/widgets.py
```diff
@@ -51,6 +51,7 @@ class Mesh(widgets.DOMWidget):
 #                                     default_value="green").tag(sync=True)
 #    geo = traitlets.Unicode('diamond').tag(sync=True)
     wire = traitlets.CBool(default_value=False).tag(sync=True)
+    visible = traitlets.CBool(default_value=True).tag(sync=True)
```
maartenbreddels: what about visible/visible_faces/visible_lines (instead of the 'wire', which is an odd name)? So the faces would be visible when `visible && visible_faces` is true, and the lines would be when `visible & visible_lines` is true. (Note that I prefer visible_xxx for tab completion, instead of xxx_visible).

js/src/mesh.js
```diff
@@ -395,6 +402,7 @@ var MeshModel = widgets.WidgetModel.extend({
             color: "red",
             sequence_index: 0,
             connected: false,
+            visible: true
```
maartenbreddels: and here visible_faces and visibles_lines as well?

ipyvolume/widgets.py
chrisjsewell: Yep no problem, I'll have a see what I can do

chrisjsewell:
random http fail, guess it just needs restarting

chrisjsewell:
FYI, I've come across a random little bug recently that I don't think was there before (at least with my setup). If try to screenshot and output the image in the same cell, then it appears in the cell where the figure is rendered:

screenshot_issue

screenshot_issue2

maartenbreddels:
Many thanks! 👍

maartenbreddels:
The bug with the screenshot may have sth to do with ipywidgets actually, and the output widget. I can reproduce it. Which ipywidgets are you using?

chrisjsewell:
7.0.0b11

PR #50, opened 2017-08-12
allow figure other than current to be screenshot

chrisjsewell:
So you can do things like this:

multi_render_example

FYI, I use these functions to do the concatenating: https://github.com/chrisjsewell/ipypublish/blob/master/ipypublish/scripts/ipynb_latex_setup.py#L90

### DIFF
--- a/ipyvolume/pylab.py
+++ b/ipyvolume/pylab.py
@@ -645,7 +645,7 @@ def _screenshot_data(timeout_seconds=10, output_widget=None, format="png", width
 
     fig.on_screenshot(screenshot_handler)
     try:
-        fig.screenshot(width=width, height=height)
+        fig.screenshot(width=width, height=height,mime_type="image/"+format)
         t0 = time.time()
         timeout = False
         ipython = IPython.get_ipython()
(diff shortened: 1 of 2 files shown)


ipyvolume/pylab.py
```diff
@@ -661,21 +664,21 @@ def screenshot_handler(image_data):
     data = data[data.find(",") + 1:]
     return base64.b64decode(data)
 
-def screenshot(timeout_seconds=10, output_widget=None, format="png", width=None, height=None):
+def screenshot(timeout_seconds=10, output_widget=None, format="png", width=None, height=None,
+               fig=None):
```
maartenbreddels: can you fix the indenting/whitespace, the last two can be on 1 line.

ipyvolume/pylab.py
```diff
@@ -661,21 +664,21 @@ def screenshot_handler(image_data):
...
     data = _screenshot_data(timeout_seconds=timeout_seconds, output_widget=output_widget,
-        format=format, width=width, height=height)
+        format=format, width=width, height=height,fig=fig)
```
maartenbreddels: ,<space>fig=fig

ipyvolume/pylab.py
```diff
@@ -661,21 +664,21 @@ def screenshot_handler(image_data):
...
     f = StringIO(data)
     return PIL.Image.open(f)
 
-def savefig(filename, timeout_seconds=10, wait=True, output_widget=None, width=None, height=None):
+def savefig(filename, timeout_seconds=10, wait=True, output_widget=None, width=None, height=None,
+           fig=None):
```
maartenbreddels: same here

ipyvolume/pylab.py
```diff
@@ -661,21 +664,21 @@ def screenshot_handler(image_data):
...
     """Save the current figure to an image (png or jpeg) to a file"""
-    # TODO: might be useful to save to a file object
     __, ext = os.path.splitext(filename)
-    fig = gcf()
     format = ext[1:]
     with open(filename, "wb") as f:
         f.write(_screenshot_data(timeout_seconds=timeout_seconds,
-            output_widget=output_widget, width=width, height=height))
+            output_widget=output_widget, width=width, height=height,fig=fig))
```
maartenbreddels: ,<space>

chrisjsewell:
I can change this if you want, but I think the [numpy docstring format](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt) is more readable than the google convention you've been using. Both are converted the same by sphinx

maartenbreddels:
I'd prefer we keep the docstring the same, it is what pycharm generates, and from this style it can infer the return type which can be useful for autocomplete. And good idea for the mime-type.

chrisjsewell:
fair, I just find them not as nice to use when I shift-tab in the notebook

maartenbreddels:
I agree actually, I may reconsider it but lets keep it consistent for now. Anyway, excellent work, thanks 👍

chrisjsewell:
No worries, FYI you can change docstring format in pycharm:  Preferences --> Tools --> Python Integrated Tools

Issue #41, opened 2017-08-02
Freezing embedded html


chrisjsewell:
Hey maarten,

As of today, I get an error when trying to view my embedded ipyvolume html files (only if it was created with the dev version):

```
ReferenceError: define is not defined
https://unpkg.com/@jupyter-widgets/html-manager@*/dist/index.js
Line 1
```

Presumably this is caused by the commit of the CombinedCamera define/import?

This was created with the pip ipyvolume (version 0.3) and conda ipywidgets (version 0.6), and still works fine:
https://chrisjsewell.github.io/ipyvolume_si_ech3.html

This was created with my dev ipyvolume (version 0.4 alpha) and pip ipywidgets (version 7.0), and no longer works:
https://chrisjsewell.github.io/3d_atomic/converted/3D%20Atomic%20Visualisation_files/ivol_example1.html

How can I rectify this and is there a way I can 'freeze' the embedded html, such that this won't happen again. I guess I need some .js files locally in the folder?

maartenbreddels:
Hi Chris,

I think https://github.com/jupyter-widgets/ipywidgets/issues/1551 was the issue, I didn't test it yet, but it should be fixed in ipywidget master, not pip yet (i guess you mean 7.0, not 0.7).
The js file for ipyvolume can be (and by default will be) copied to the same folder as the html file. If ipyvolume.js cannot be found, it will be taken from unpkg.com, with the right version number. But I think now the main problem is ipywidgets. I hope I can test that out.

cheers,

Maarten

chrisjsewell:
Ok panic over, in jupyter-widgets/ipywidgets#1556 the following has been changed:

`DEFAULT_EMBED_SCRIPT_URL = u'https://unpkg.com/@jupyter-widgets/html-manager@*/dist/index.js'`

to 

DEFAULT_EMBED_SCRIPT_URL = u'https://unpkg.com/@jupyter-widgets/html-manager@*/dist/embed.js'

So the `ipyvolume.embed.embed_html` will technically be faulty unless used with ipywidgets versions after this commit.

chrisjsewell:
But obviously the more general desire remains; to have a static/offline ipyvolume rendering that isn't dependent on (and can't be broken by) any external sources.

chrisjsewell:
Also `ipyvolume.embed.embed_html` doesn't appear to be outputting ipyvolume.js any more. I recall this being fine in version 0.3, has anything changed?

[...]

PR #55, opened 2017-08-17
Move embedding logic to IPywidgets and add an offline option

chrisjsewell:
As discussed in maartenbreddels/ipyvolume#41

This is all working in my python 3.6 environment, except for one main issue: the embedded HTML no longer works with index.js -> ipyvolume.js from ipyvolume/static (the notebook version), only the unpkg version from js/dist 

The only other little annoyance is that embed-amd.js requires that ipyvolume.js be in the same folder as the embedded HTML, rather than the same folder as itself. Since I'd like to have them like this:

    my_embedded.html
    scripts_folder
        embed-amd.js
        ipyvolume.js
        font-awesome
        require.min.js

@jasongrout, I don't know if there is a better way to get hold of the embed-amd.js file?

### DIFF
--- a/ipyvolume/embed.py
+++ b/ipyvolume/embed.py
@@ -5,6 +5,7 @@ import shutil
 from ipywidgets import embed as wembed
 import ipyvolume
 from ipyvolume.utils import download_to_file, download_to_bytes
+from ipyvolume._version import __version_threejs__
 
 html_template = u"""<!DOCTYPE html>
 <html lang="en">
@@ -22,17 +23,17 @@ html_template = u"""<!DOCTYPE html>
 """
 
 
-def save_ipyvolumejs(folderpath="", version=ipyvolume._version.__version_js__, devmode=False):
+def save_ipyvolumejs(target="", version=ipyvolume._version.__version_js__, devmode=False):
     """ output the ipyvolume javascript to a local file
     
-    :type folderpath: str
+    :type target: str
     :type version: str
     :type devmode: bool
 
     """
     url = "https://unpkg.com/ipyvolume@{version}/dist/index.js".format(version=version)
     pyv_filename = 'ipyvolume_v{version}.js'.format(version=version)
-    pyv_filepath = os.path.join(folderpath, pyv_filename)
+    pyv_filepath = os.path.join(target, pyv_filename)
 
     devfile = os.path.join(os.path.abspath(ipyvolume.__path__[0]), "..", "js", "dist", "index.js")
     if devmode and os.path.exists(devfile):
@@ -50,60 +51,60 @@ def save_ipyvolumejs(folderpath="", version=ipyvolume._version.__version_js__, d
     return pyv_filename, three_filename
 
 
-def save_requirejs(folderpath="", version="2.3.4"):
+def save_requirejs(target="", version="2.3.4"):
     """ download and save the require javascript to a local file
 
-    :type folderpath: str
+    :type target: str
     :type version: str
     """
     url = "https://cdnjs.cloudflare.com/ajax/libs/require.js/{version}/require.min.js".format(version=version)
     filename = "require.min.v{0}.js".format(version)
-    filepath = os.path.join(folderpath, filename)
+    filepath = os.path.join(target, filename)
     download_to_file(url, filepath)
     return filename
 
 
-def save_embed_js(folderpath="", version=wembed.__html_manager_version__):
+def save_embed_js(target="", version=wembed.__html_manager_version__):
     """ download and save the ipywidgets embedding javascript to a local file
 
-    :type folderpath: str
+    :type target: str
     :type version: str
 
     """
     url = u'https://unpkg.com/@jupyter-widgets/html-manager@{0:s}/dist/embed-amd.js'.format(version)
     filename = "embed-amd_v{0:s}.js".format(version[1:])
-    filepath = os.path.join(folderpath, filename)
+    filepath = os.path.join(target, filename)
 
     download_to_file(url, filepath)
     return filename
 
 
 def save_font_awesome(dirpath='', version="4.7.0"):
-    """ download and save the font-awesome package to a local folder
+    """ download and save the font-awesome package to a local directory
 
     :type dirpath: str
     :type url: str
 
     """
-    folder_name = "font-awesome-{0:s}".format(version)
-    folder_path = os.path.join(dirpath, folder_name)
-    if os.path.exists(folder_path):
-        return folder_name
+    directory_name = "font-awesome-{0:s}".format(version)
+    directory_path = os.path.join(dirpath, directory_name)
+    if os.path.exists(directory_path):
+        return directory_name
 
     url = "http://fontawesome.io/assets/font-awesome-{0:s}.zip".format(version)
     content, encoding = download_to_bytes(url)
 
     try:
-        zip_folder = io.BytesIO(content)
-        unzip = zipfile.ZipFile(zip_folder)
+        zip_directory = io.BytesIO(content)
+        unzip = zipfile.ZipFile(zip_directory)
         top_level_name = unzip.namelist()[0]
         unzip.extractall(dirpath)
     except Exception as err:
         raise IOError('Could not unzip content from: {0}\n{1}'.format(url, err))
 
-    os.rename(os.path.join(dirpath, top_level_name), folder_path)
+    os.rename(os.path.join(dirpath, top_level_name), directory_path)
 
-    return folder_name
+    return directory_name
 
 
 def embed_html(filepath, widgets, makedirs=True, title=u'IPyVolume Widget', all_states=False,
@@ -122,12 +123,12 @@ def embed_html(filepath, widgets, makedirs=True, title=u'IPyVolume Widget', all_
     :param all_states: if True, the state of all widgets know to the widget manager is included, else only those in widgets
     :param offline: if True, use local urls for required js/css packages and download all js/css required packages
     (if not already available), such that the html can be viewed with no internet connection
-    :param scripts_path: the folder to save required js/css packages to (relative to the filepath)
+    :param scripts_path: the directory to save required js/css packages to (relative to the filepath)
     :type drop_defaults: bool
     :param drop_defaults: Whether to drop default values from the widget states
     :param template: template string for the html, must contain at least {title} and {snippet} place holders
     :param template_options: list or dict of additional template options
-    :param devmode: if True, attempt to get index.js from local js/dist folder
+    :param devmode: if True, attempt to get index.js from local js/dist directory
     :param cors: if True, sets crossorigin attribute to anonymous
 
     """
--- a/ipyvolume/embed.py
+++ b/ipyvolume/embed.py
@@ -77,7 +77,7 @@ def save_font_awesome(dirpath='font-awesome', url="http://fontawesome.io/assets/
 
     """
     parentdirname = os.path.dirname(dirpath)
-    print(print("Downloading %s to %s" % (url, os.path.abspath(dirpath))))
+    print("Downloading %s to %s" % (url, os.path.abspath(dirpath)))
     try:
         zip_folder = io.BytesIO(requests.get(url).content)
         unzip = zipfile.ZipFile(zip_folder)
--- a/ipyvolume/embed.py
+++ b/ipyvolume/embed.py
@@ -149,6 +149,8 @@ def embed_html(filepath, widgets, makedirs=True, title=u'IPyVolume Widget', all_
         shutil.copy(threejs, directory)
     else:
         if offline_req:
+            if not os.path.isabs(scripts_path):
+                scripts_path = os.path.join(os.path.dirname(filepath), scripts_path)
             # ensure script path is above filepath
             rel_script_path = os.path.relpath(scripts_path, os.path.dirname(filepath))
             if rel_script_path.startswith(".."):
--- a/ipyvolume/embed.py
+++ b/ipyvolume/embed.py
@@ -1,132 +1,15 @@
 import os
-import json
+import io
+import requests
+import zipfile
 import shutil
-import ipywidgets
+from ipywidgets import embed as wembed
 import ipyvolume
-from base64 import standard_b64encode
-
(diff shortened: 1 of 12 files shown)


ipyvolume/embed.py
```diff
@@ -140,63 +57,118 @@ def save_js(dirname, makedirs=True):
...
+        zip_folder = io.BytesIO(requests.get(url).content)
+        unzip = zipfile.ZipFile(zip_folder)
+        top_level_name = unzip.namelist()[0]
+        unzip.extractall(parentdirname)
+    except Exception as err:
+        raise IOError('Could not save: {0}\n{1}'.format(url, err))
+
+    if os.path.exists(dirpath):
+        shutil.rmtree(dirpath)
+    os.rename(os.path.join(parentdirname, top_level_name), dirpath)
```
maartenbreddels: This should be included by webpack file (see https://github.com/maartenbreddels/ipyvolume/issues/41#issuecomment-322916169) , can you create an issue for this on ipywidgets? If fixed, we can delete this.

ipyvolume/embed.py
```diff
@@ -140,63 +57,118 @@ def save_js(dirname, makedirs=True):
     src = os.path.join(dir_name_src, "index.js")
     shutil.copy(src, dst)
```
maartenbreddels: You'r right, it's not index.js in the static dir. I think we can do this smart. If we are in dev mode (ipywidgest module is a symlink), the like should be at <ipywidgets.__file__>../js/dist/index.js if that directory doesn't exists, we are not in dev mode and we should fetch it online. The other option would be to include the file (say ipyvolume.js in the same static folder as well).

ipyvolume/embed.py
```diff
@@ -140,63 +57,118 @@ def save_js(dirname, makedirs=True):
...
+def save_requirejs(filepath='require.min.js', url="https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.4/require.min.js"):
+    """ download and save the require javascript to a local file
+
+    :type filepath: str
+    :type url: str
+    """
+    content = requests.get(url).content
+    with io.open(filepath, 'w', encoding='utf8') as f:
+        f.write(content.decode("utf8"))
+
```
maartenbreddels: Maybe we should have some .download(url, path) function in utils.py (see also https://github.com/maartenbreddels/ipyvolume/blob/master/ipyvolume/datasets.py#L35). Also note that requests is not a stdlib, urlretrieve is in the stdlib.

ipyvolume/embed.py
```diff
@@ -140,63 +57,118 @@ def save_js(dirname, makedirs=True):
...
+    :type filepath: str
+    :type url: str
+    """
+    if not url.endswith('.js'):
+        url += '.js'
+    content = requests.get(wembed.DEFAULT_EMBED_REQUIREJS_URL).content
+    with io.open(filepath, 'w', encoding='utf8') as f:
+        f.write(content.decode("utf8"))
+
```
maartenbreddels: so this can be ipyvolume.utils.download(...)

ipyvolume/embed.py
maartenbreddels: cc @jasongrout
If we want this to go upstream, we need to do this differently. Scan the while state for all module names, and versions, and fetch them from a CDN, and run require.config() so it knows where to fetch them. Then there is the issue with multiple versions of the same package.. but maybe that's not sth we want to support initially.

ipyvolume/embed.py
```diff
@@ -140,63 +57,118 @@ def save_js(dirname, makedirs=True):
...
+        offline_snippet = """
+        <link href="{offline_folder}font-awesome/css/font-awesome.min.css" rel="stylesheet">    
+        <script src="{offline_folder}require.min.js" crossorigin="anonymous"></script>
+        {snippet}
+        """.format(offline_folder=offline_folder, snippet=snippet)
+
+        template_opts['snippet'] = offline_snippet
+
+    fmt = DefaultFormatter()
+    html_code = fmt.format(template, **template_opts)
```
maartenbreddels: I'd keep the {}.update() strategy. Less people would know what the defaultformatter does, would have to look it up etc, I'd say it goes agains Python zen :smile:

ipyvolume/embed.py
```diff
@@ -140,63 +57,118 @@ def save_js(dirname, makedirs=True):
...
+    :param filepath: The file to write the HTML output to.
+    :type widgets: widget or collection of widgets or None
+    :param widgets:The widgets to include views for. If None, all DOMWidgets are included (not just the displayed ones).
+    :param makedirs: whether to make directories in the filename path, if they do not already exist
+    :param title: title for the html page
+    :param all_states: if True, the state of all widgets know to the widget manager is included, else only those in widgets
+    :param offline: if True, use local urls for required js/css packages
+    :param offline_req: if True and offline=True, download all js/css required packages,
+    such that the html can be viewed with no internet connection
+    :param offline_folder: the folder to save required js/css packages to (relative to the filepath)
```
maartenbreddels: Maybe script_directory?

ipyvolume/pylab.py
```diff
@@ -585,10 +585,11 @@ def volshow(data, lighting=False, data_min=None, data_max=None, tf=None, stereo=
     return vol
 
 
-def save(filename, copy_js=True, makedirs=True, **kwargs):
-    """Save the figure/visualization as html file, and optionally copy the .js file to the same directory """
-    ipyvolume.embed.embed_html(filename, current.container,
-                       makedirs=makedirs, copy_js=copy_js, **kwargs)
+def save(filename, makedirs=True, offline=False, offline_req=True, **kwargs):
+    """Save the figure/visualization as html file, and optionally copy the required .js files to the same directory """
```
maartenbreddels: see other comments.

setup.py
```diff
@@ -137,7 +137,8 @@ def run(self):
         'traittypes',
         'traitlets',
         'Pillow',
-        'ipywebrtc'
+        'ipywebrtc',
+        'requests'
```
maartenbreddels: Can we do without (see other comment)

chrisjsewell: I would point out that the python documentation directly recommends the request package: https://docs.python.org/3/library/urllib.request.html. But I can change to `urllib.request.urlretrieve` if you want?

ipyvolume/embed.py
chrisjsewell: Is there a reason in https://github.com/maartenbreddels/ipyvolume/blob/master/ipyvolume/datasets.py#L35 that you try wget and curl first?

setup.py
maartenbreddels: Ok, I leave it up to you, but the less dependencies the better I'd think. And I like that urlretrieve is a oneliner (except for the import).

ipyvolume/embed.py
maartenbreddels: Yes, they are mature, and would do better in downloading large files, have automatic resume, and being a separate process you're sure it's not a memleak. With wget you also get a progressbar.

jasongrout:
> The only other little annoyance is that embed-amd.js requires that ipyvolume.js be in the same folder as the embedded HTML, rather than the same folder as itself. Since I'd like to have them like this:

This is how requirejs works. It has no idea where to find the ipyvolume.js, and doesn't assume that it is the same as the embedder.

One thing you could do is declare a requirejs path somewhere that specifically tells require where to load it, using a require.config call. That config would need to happen up at the top of the page before the embed-amd include.

maartenbreddels:
Indeed, require.config would be the way to go, you'll lose support for having multiple versions of the same module at the same page (see comment above), although that would be a rare case I'd think.

ipyvolume/embed.py
chrisjsewell: done, added devmode option to switch on fetching from js/dist/index.js so that, when I can turn this off when running tests

ipyvolume/embed.py
chrisjsewell: done. I've kept with the `requests` package, because its definitely better the urlretrieve and its allowed me to get pretty good (I think), pythonic version of the wget/curl feature: `ipyvolume.utils.download_to_file` is able to download in chunks, resume from partially downloaded files and, better than wget/curl, has a simple progress bar when running in both the console and notebook. Also `download_to_bytes` and `download_yield_bytes` allow files to be downloaded in memory. Give it a go, see what you think 😀

ipyvolume/embed.py
chrisjsewell: done (script_path)

ipyvolume/embed.py
chrisjsewell: done

maartenbreddels:
@chrisjsewell  Would you mind if I push some changes to your branch?

chrisjsewell:
I guess it depends what they are and if you're going to undo all my good work lol?

maartenbreddels:
No, some things were not working, for instance offline=False make an empty html, and we now need to copy three.js as well. Also the script_path can be relative now.

chrisjsewell:
Yeh oops thats a bit of a major bug! Yeh no that's fine obviously, thanks for asking

chrisjsewell:
the only thing I had left to do is change ipyvolume.js to be {script_path}/ipyvolume_v{versions}.js and add the require config script to the offline snippet

maartenbreddels:
still sth odd going on, embed sometimes outputs all (non visible) widgets.

chrisjsewell:
Is this for offline True and False, do you think it's to do with `state = wembed.dependency_state(widgets, drop_defaults=drop_defaults)`?

chrisjsewell:
Don't know if this is what you were referring to, but `test_embed_offline` was inheriting widgets from `test_quick`, so I've added a `p3.clear()`

chrisjsewell:
is the inclusion of three.js for a future update? Because at present its not required, and so I can test if it is working correctly

maartenbreddels:
I don't understand name 'offline_req` nor it's indented behaviour. I expect that if I pass offline=True, that it will download it, if not already.

chrisjsewell:
That why I got rid of it 😀

maartenbreddels:
Just noticed, I had a merge conflict :).
I just had it again, instead of 1 plot, I had like 10, but still no idea what is causing this. I think we're nearly there.

ipyvolume/embed.py
```diff
@@ -1,202 +1,199 @@
...
+    pyv_filename = 'ipyvolume_v{version}.js'.format(version=version)
+    pyv_filepath = os.path.join(folderpath, pyv_filename)
 
-<script src="https://unpkg.com/jupyter-js-widgets@~2.0.20/dist/embed.js"></script>
-<script type="javascript">
-widget_data = {json_data}
-</script>
-{widget_views}
+    devfile = os.path.join(os.path.abspath(ipyvolume.__path__[0]), "..", "js", "dist", "index.js")
+    if devmode and os.path.exists(devfile):
```
maartenbreddels: Not sure about this, is devmode=True, make it crash if it cannot find it, silently passing might be confusing.

ipyvolume/embed.py
```diff
@@ -1,202 +1,199 @@
...
+        if folderpath and not os.path.exists(folderpath):
```
maartenbreddels: Do you mind if I do a s/folder/directory ?

ipyvolume/embed.py
```diff
@@ -1,202 +1,199 @@
...
+            os.makedirs(folderpath)
+        shutil.copy(devfile, pyv_filepath)
+    else:
+        download_to_file(url, pyv_filepath)
 
-{body_post}
-</body>
-</html>
-"""
+    three_filename = 'three_v{version}.js'.format(version=version)
```
maartenbreddels: threejs has it's own version, this might be confusing. The threejs version also won't change often, so please just keep it three.js, or we should add the threejs version somewhere in this file (you can find the version in js/modules.json.

ipyvolume/embed.py
```diff
@@ -1,202 +1,199 @@
...
     """
+    url = u'https://unpkg.com/@jupyter-widgets/html-manager@{0:s}/dist/embed-amd.js'.format(version)
+    filename = "embed-amd_v{0:s}.js".format(version[1:])
+    filepath = os.path.join(folderpath, filename)
+
+    download_to_file(url, filepath)
+    return filename
+
+
+def save_font_awesome(dirpath='', version="4.7.0"):
```
maartenbreddels: could you mark this as TODO, and link to the original issue on ipywidgets. I'm happy to keep this for now though.

ipyvolume/utils.py
```diff
@@ -16,4 +21,163 @@ def nested_setitem(obj, dotted_name, value):
...
+                        loaded += 1
+                    loaded_size += chunk_size
+                f.write(chunk)
+        if total_length is None:
+            print("=" * loadbar_length, end='')
+        else:
+            while loaded < loadbar_length:
+                print("=", end='')
+                loaded += 1
+    print("] Finished")
```
maartenbreddels: You should make this into a separate library, I like it :)

maartenbreddels:
I'd like to do a new release soon, so I did some of this myself. Do you mind if I merge?

PS: you were right about the three.js part not beeing needed, that was only for in the notebook, but it might happen in the future, so I'm keeping some of it in.

chrisjsewell:
Ah sorry I have made all the changes, except s/folder/directory. You do have an issue though... The latest embedded html (including those at http://ipyvolume.readthedocs.io/en/latest/) is failing to load in Safari on OSX and iOS (and Chrome and Firefox on iOS) with the same error:

image

maartenbreddels:
Hey, sorry that it went this way, but I need to merge this (excellent work) fast, since I want to have the release out before JupyterCon. And this way I can go over it while traveling.
Safari may have to wait, I'll see what I can do.

maartenbreddels:
Do you mind if I merge this?

chrisjsewell:
Hey yeh no problem :)

maartenbreddels:
thanks for the great work 👍

chrisjsewell:
cheers :) Do you want me to open an issue about Safari?

maartenbreddels:
Yes, that would be nice, so we don't lose track of it.

PR #63, opened 2017-09-05
binder link fix

choldgraf:
This fixes the link so it works w/ new binder...the requirements etc didn't need to change at all.


maartenbreddels:
Thanks! The travis failure is unrelated.

PR #65, opened 2017-09-07
adding docstring for animation

choldgraf:
adds an extra sentence to a few of your docstrings...lemme know if you'd like a different wording or something

### DIFF
--- a/ipyvolume/pylab.py
+++ b/ipyvolume/pylab.py
@@ -25,6 +25,8 @@ def _docsubst(f):
     f.__doc__ = f.__doc__.format(**_doc_snippets)
     return f
 
+_seq_sn = "If an (S, N) array, the first dimension will be used for frames in an animation."
+_seq_snm = "If an (S, N, M) array, the first dimension will be used for frames in an animation."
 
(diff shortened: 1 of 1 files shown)


maartenbreddels:
Thanks for attacking this.
My 'plan' was to add the snippet `(can be a sequence)`, and somewhere on the top explain that this can be an animation, but that is too abstract and requires you to read more than just that docstring, which is bad. What about:
```
numpy array of shape (N,) or (S, N) with x positions, when of shape (S, N), the first dimension will be used for frames in an animation.
```
And similar for the rest?

choldgraf:
sounds fine to me - that's pretty close to the wording I've got here as well. I'll use your wording, though will use a period in between the first part of the docstring and this part so that it's a little more modular...cool?

maartenbreddels:
yeah, basically remove the 'can be a sequence', and then I'll trust you on the proper language 👍

choldgraf:
ok so how is this?

maartenbreddels:
Excellent, thanks 👍

choldgraf:
woo!

Issue #74, opened 2017-10-07
Other transfer functions


OliverEvans96:
Hi @maartenbreddels,

I remember you expressing interest in other transfer functions. Currently, it seems that the TF with RGB gaussians is the only one implemented. I see that there are a few TransferFunction objects defined at https://github.com/maartenbreddels/ipyvolume/blob/master/ipyvolume/transferfunction.py.

I'm wondering if there's a clear entrypoint for implementing other transfer functions. Is it via modifying these objects I pointed out, or would some other infrastructure need to be rearranged? Can it be done via Python alone, or will it require JS?

Thanks!
Oliver

maartenbreddels:
Hi Oliver,

The current implementation that is used is that of [TransferFunctionWidgetJs3][https://github.com/maartenbreddels/ipyvolume/blob/771b76987e03800ddafa558f0518546d85b8094b/ipyvolume/transferfunction.py#L40] where all the work is being done on the [js/browser side](https://github.com/maartenbreddels/ipyvolume/blob/771b76987e03800ddafa558f0518546d85b8094b/js/src/tf.js#L37).
However, the only thing that needs to be done is set the [rgba trait of it's parent class (TransferFunction)](https://github.com/maartenbreddels/ipyvolume/blob/771b76987e03800ddafa558f0518546d85b8094b/ipyvolume/transferfunction.py#L31).
You should be able to read out the rgba value of the Python side for the JS implementation, and assign this to a newly created transfer function, and play around with that. To be honoust I haven't tested this for a while, feel free to raise issues on this, happy to give extra pointers or do small fixes.

cheers,

Maarten

maartenbreddels:
Just telling you know i'm testing things out, and transferring the `.rgba` trait is broken, so don't spend too much time on this yet :)

maartenbreddels:
Ok, some fixes in master now.

Starting with a 3d plot like this:
```python
import ipyvolume as ipv
import ipywidgets
import numpy as np
ds = ipv.datasets.aquariusA2.fetch()
fig = ipv.figure()
ipv.volshow(ds.data)
ipv.show()
```
We can now visualize the transfer function as image:
```python
import PIL.Image
# * 255 for [0, 1] to [0,256) range
rgba = fig.tf.rgba * 255  # optionally multiply by ~4 to see the colors better
rgba = (np.repeat([rgba], 256, axis=0)) # repeat so we get a height of 256, instead of 1 pixel
print(rgba.shape)
[...]

PR #76, opened 2017-10-13
Add TransferFunction editing notebook

OliverEvans96:
Create a widget interface with bqplot's HandDraw to interactively edit TransferFunctions, as discussed in #74 .

image

### DIFF
--- a/ipyvolume/widgets.py
+++ b/ipyvolume/widgets.py
@@ -161,6 +161,10 @@ class Figure(ipywebrtc.MediaStream):
     width = traitlets.CInt(500).tag(sync=True)
     height = traitlets.CInt(400).tag(sync=True)
     downscale = traitlets.CInt(1).tag(sync=True)
+    displayscale = traitlets.CFloat(1).tag(sync=True)
+    capture_fps = traitlets.CFloat(None, allow_none=True).tag(sync=True)
+    cube_resolution = traitlets.CInt(512).tag(sync=True)
+
     show = traitlets.Unicode("Volume").tag(sync=True) # for debugging
 
     xlim = traitlets.List(traitlets.CFloat, default_value=[0, 1], minlen=2, maxlen=2).tag(sync=True)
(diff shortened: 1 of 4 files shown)


maartenbreddels:
I tried it out, and it's really nice! I like the idea of using bqplot. Do you have any more plans with this?

OliverEvans96:
That was the extent of my plans so far for this widget. I thought it would
be helpful in order to visualize and edit transfer functions.

Ultimately, it would be nice to have a few classes of transfer functions
ready to use for ipyvolume that could be easily chosen with a kwarg to
volshow, perhaps.

Do you have any thoughts or suggestions? About the theory of transfer
functions or about how to use this type of interactive widget?

Oliver

maartenbreddels:
What I'd like to see, is an instant change of the transfer function, I think that requires a change in bqplot's handdrawn widget. I'm thinking of making this into an example to put on the documentation page, or something like a 'contribute' subpackage where this could go so other people might use it. But given the discussion in #74 it might be too early :)

OliverEvans96:
What do you mean by that? As in you change the sliders in quickvolshow and
the rgba curves update automatically? If so, I thought about this. I put
the Reload TF button to do this manually, but if I were to have this happen
automatically, I think it would cause infinite recursion: changing the
curves changes the TF which changes the curves, etc.

If the two were linked with traitlets.link, this wouldn't be an issue, but
I'm currently using observe to set the TF since I'm combining data from 4
curves into one numpy rgba array. If you have any thoughts about this, let
me know!

If you meant something else by an instant change in the TF, then I'm not
sure I understood.

I'd be glad to change things around if you'd like so that it would make for
a good example.

maartenbreddels:
When you draw by hand, it only updates when you let the mouse button go (at least here), would be nice if every mouse change it would update the rgba values.

There is also traitlets.dlink/ipywidgets.dlink (which does a directional like, and ipwidgets' one sets it on the js side).

OliverEvans96:
Ah, yes - I see what you mean. I agree, that might require modifying
bqplot. But it seems reasonable that there should be some analogue of
continupus_update for handdraw. I'm thinking this should be doable.

OliverEvans96:
I had a thought for the widget: it would be great to specify other color
spaces. In particular, specifying hsl/alpha would be neat.

I found  another idea I thought made a lot of sense in this tutorial:
http://graphicsrunner.blogspot.com/2009/01/volume-rendering-102-transfer-functions.html?m=1

In their TF editing widget, you specify the endpoints of a horizontal
linear colorbar, and then have a single alpha curve to edit as a function
of data value.
When I've explained the current system with the 3 RGB gaussians to my
professors, it's always taken a minute for them to understand that red is
low, green is medium, and blue is high. I think having a single linear
color gradient from low to high could eliminate that confusion.

maartenbreddels:
This is exactly the rendering I want to be able to reproduce! I lost the link, so thanks for putting it here. This person also has some interesting ideas for optimization, really worth a look.

I actually had an implementation with a color gradient (all the colormaps from matplotlib) and it is not really great. If your 'alpha bumps' are near eachother, the will also have a quite similar color, and are impossible to distinguish. The red, green, blue sequence for me is easy to remember :)

PR #86, opened 2017-11-25
Mesh generation speed

xavArtley:
Modifcation of the function to create triangles and lines of mesh plots
Mesh creation is 100x more quick
mesh_2000x2000

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,10 +75,8 @@ Or with anaconda/conda:
 
 For a development installation (requires npm),
 
-    $ git clone https://github.com/maartenbreddels/ipyvolume.git	
-    $ cd ipyvolume/js
-	$ nmp install
-	$ cd ..
+    $ git clone https://github.com/maartenbreddels/ipyvolume.git
+    $ cd ipyvolume
     $ pip install -e .
     $ jupyter nbextension install --py --symlink --sys-prefix ipyvolume
     $ jupyter nbextension enable --py --sys-prefix ipyvolume
--- a/README.md
+++ b/README.md
@@ -75,8 +75,10 @@ Or with anaconda/conda:
 
 For a development installation (requires npm),
 
-    $ git clone https://github.com/maartenbreddels/ipyvolume.git
-    $ cd ipyvolume
+    $ git clone https://github.com/maartenbreddels/ipyvolume.git	
+    $ cd ipyvolume/js
+	$ nmp install
+	$ cd ..
     $ pip install -e .
     $ jupyter nbextension install --py --symlink --sys-prefix ipyvolume
     $ jupyter nbextension enable --py --sys-prefix ipyvolume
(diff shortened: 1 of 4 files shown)


maartenbreddels:
Nice work, will look at this soon, I'll do a small review now.

README.md
```diff
@@ -75,8 +75,10 @@ Or with anaconda/conda:
 
 For a development installation (requires npm),
 
-    $ git clone https://github.com/maartenbreddels/ipyvolume.git
-    $ cd ipyvolume
+    $ git clone https://github.com/maartenbreddels/ipyvolume.git	
+    $ cd ipyvolume/js
+	$ nmp install
+	$ cd ..
```
maartenbreddels: No need to do the npm install manually, the setup.py should do this, or did it somehow give you issues?

ipyvolume/pylab.py
```diff
@@ -1007,3 +979,47 @@ def lasso(data, other=None, fig=fig):
                     if selected:
                         scatter.selected = selected
     fig.on_lasso(lasso)
+    
+
+
+def make_triangles_lines(x, y, z, wrapx=False, wrapy=False):
```
maartenbreddels: could you put a _ in front, so signal this as private api?

ipyvolume/pylab.py
```diff
@@ -1007,3 +979,47 @@ def lasso(data, other=None, fig=fig):
...
+    assert len(z.shape) == 2, "Array z must be 2 dimensional."
+    assert x.shape == y.shape, "Arrays x and y must have same shape."
+    assert y.shape == z.shape, "Arrays y and z must have same shape."
+    
+    nx, ny = x.shape
+
+    mx = nx if wrapx else nx - 1
+    my = ny if wrapy else ny - 1
+    
+    i, j = np.mgrid[0:mx, 0:my]
```
maartenbreddels: from this part on it's a bit difficult to follow, could you put in some inline comments to explain the idea? For instance give an example for nx,ny = (2,3) orso.

README.md
xavArtley: I really don't understand how npm and javascript works but when I tried to install using setup.py it failed.
I found this solution running npm before to create javascript lib before running setup

maartenbreddels: Ok, that is odd, could you open an issue for that and remove this part?

ipyvolume/pylab.py
xavArtley: I'll will look at some explanation this week-end, but the function does exactly the same things as your function but without using for loops wich are really slow in python

maartenbreddels: I understand, and I don't think it's gonna be a problem to decypher it, but for maintainance etc it would be nice to explain it a bit.

maartenbreddels:
I've manually merged this PR. But the author information doesn't match your github account. I've changed your author name, but I think also the email should match the email you use for your account.

maartenbreddels:
Solved, the commit is here:
a923744d6d754eea314a5405c996cfd702a0733b

PR #103, opened 2018-01-27
Fix compatibility with Python 2

lrq3000:
This PR fixes a compatibility issue with Python 2: decorating with just `@widgets.register` will transform the decorated class into a function, whereas using `@widgets.register()` will correctly return a class. I could successfully run the example:

```python
import ipyvolume
ipyvolume.examples.ball(rmax=3, rmin=2.5, shape=64, lighting=True)
```

BTW it would be nice I think to add the [mybinder examples](https://hub.mybinder.org/user/maartenbreddels-ipyvolume-ew8oywwc/notebooks/notebooks/simple.ipynb) in the doc directly, or even the readme, so that new users can easily test out the widget for themselves :-) Otherwise in the doc there is no other example that does not require a 3rd-party library (bokeh etc).

### DIFF
--- a/ipyvolume/transferfunction.py
+++ b/ipyvolume/transferfunction.py
@@ -28,7 +28,7 @@ if 0:
 
 	array_serialization = dict(to_json=array_to_json, from_json=array_from_json)
 
-@widgets.register
+@widgets.register()
 class TransferFunction(widgets.DOMWidget):
 	_model_name = Unicode('TransferFunctionModel').tag(sync=True)
 	_view_name = Unicode('TransferFunctionView').tag(sync=True)
--- a/ipyvolume/widgets.py
+++ b/ipyvolume/widgets.py
@@ -20,7 +20,7 @@ _last_volume_renderer = None
 import ipyvolume._version
 semver_range_frontend = "~" + ipyvolume._version.__version_js__
 
-@widgets.register
+@widgets.register()
 class Mesh(widgets.DOMWidget):
     _view_name = Unicode('MeshView').tag(sync=True)
     _view_module = Unicode('ipyvolume').tag(sync=True)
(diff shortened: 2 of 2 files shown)


lrq3000:
Also for reference (about the decorator):
* https://bugs.python.org/issue6829

maartenbreddels:
Hi there, thanks for your contribution!
I cannot reproduce this actually.
The decorator might be a bit confusing, since going from ipywidget 6 to 7 it does not take a name anymore, but a widget instead. If you installed ipyvolume with ipywidget 6 this may happen actually!.
A binder link in the docs would be great, I saw some auto generated once, I'll look into it.

lrq3000:
@maartenbreddels mmm that's weird, did you try with Python >= 2.7.13 ? Because it's known to work with Python >= 3 since the class management changed, so this issue affects only Python 2. I have ipywidget v7.1.1, I had to update else the widget would not even load with the time old issue of "javascript can't be loaded" or something like that.

maartenbreddels:
it's 2.7.14, on OSX, using miniconda, and ipyvolume works with that. I'm almost sure it is ipywidget 6 you are using, maybe you have it installed in `~/.local`?
can you check the output of `python -c "import ipywidgets as w; print(w.__version__, w.__file__)"` ?

lrq3000:
Hey @maartenbreddels , thank you for your suggestion, here is the output:

```python
('7.1.1', 'C:\\Anaconda2\\lib\\site-packages\\ipywidgets\\__init__.pyc')
```

The path is where I initially applied my patch to make ipyvolume works. Also my Python version is 2.7.12, maybe there was an update between .12 and .14 that renders Py2 compatible with such decorator?

maartenbreddels:
Ok, I need to check this on windows, I only have OSX and Linux now, but I'll try to boot up a virtual machine soon!

PR #104, opened 2018-02-07
Headless fixes

mathieu1:
The headless example is currently broken, and requires a few fixes:

- [x] The font-awesome url is no longer valid
- [x] Do not force `devmode=True` in headless mode
- [x] ~Fetch some version of `three.js`~
- [x] Do not use offline mode when headless

### DIFF
--- a/ipyvolume/embed.py
+++ b/ipyvolume/embed.py
@@ -102,8 +102,7 @@ def save_font_awesome(dirpath='', version="4.7.0"):
     directory_path = os.path.join(dirpath, directory_name)
     if os.path.exists(directory_path):
         return directory_name
-
-    url = "http://fontawesome.io/assets/font-awesome-{0:s}.zip".format(version)
+    url = "https://fontawesome.com/v{0:s}/assets/font-awesome-{0:s}.zip".format(version)
     content, encoding = download_to_bytes(url)
 
     try:
--- a/ipyvolume/pylab.py
+++ b/ipyvolume/pylab.py
@@ -754,7 +754,7 @@ def _screenshot_data(timeout_seconds=10, output_widget=None, format="png", width
         import tempfile
         tempdir = tempfile.mkdtemp()
         tempfile = os.path.join(tempdir, 'headless.html')
-        save(tempfile, offline=True, scripts_path=tempdir, devmode=devmode)
+        save(tempfile, offline=False, scripts_path=tempdir, devmode=devmode)
         data = headless._screenshot_data("file://" + tempfile)
     else:
         if output_widget is None:
--- a/ipyvolume/pylab.py
+++ b/ipyvolume/pylab.py
@@ -744,7 +744,7 @@ def movie(f="movie.mp4", function=_change_y_angle, fps=30, frames=30, endpoint=F
     return tempdir
 
 
-def _screenshot_data(timeout_seconds=10, output_widget=None, format="png", width=None, height=None, fig=None, headless=False):
+def _screenshot_data(timeout_seconds=10, output_widget=None, format="png", width=None, height=None, fig=None, headless=False, devmode=False):
     if fig is None:
         fig = gcf()
     else:
@@ -754,7 +754,7 @@ def _screenshot_data(timeout_seconds=10, output_widget=None, format="png", width
         import tempfile
         tempdir = tempfile.mkdtemp()
         tempfile = os.path.join(tempdir, 'headless.html')
-        save(tempfile, offline=True, scripts_path=tempdir, devmode=True)
+        save(tempfile, offline=True, scripts_path=tempdir, devmode=devmode)
         data = headless._screenshot_data("file://" + tempfile)
     else:
         if output_widget is None:
@@ -792,7 +792,7 @@ def _screenshot_data(timeout_seconds=10, output_widget=None, format="png", width
     data = data[data.find(",") + 1:]
     return base64.b64decode(data)
 
-def screenshot(width=None, height=None, format="png", fig=None, timeout_seconds=10, output_widget=None, headless=False):
+def screenshot(width=None, height=None, format="png", fig=None, timeout_seconds=10, output_widget=None, headless=False, devmode=False):
     """Save the figure to a PIL.Image object.
 
     :param int width: the width of the image in pixels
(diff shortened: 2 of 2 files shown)


mathieu1:
The last fix (fetching `three.js`) feels quite hacky, here are a few reasons why:

1. The generated html looks like this (with or without the fix):
```html
<head>
...
<script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="ipyvolume_v0.4.5" src="./ipyvolume_v0.4.5.js"></script>
<script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="three" src="./three.js"></script>
<script type="text/javascript" charset="utf-8" async="" data-requirecontext="_" data-requiremodule="https://unpkg.com/ipyvolume@~0.4.5/dist/index.js" src="https://unpkg.com/ipyvolume@~0.4.5/dist/index.js"></script>
...
</head>
<body>
...
<script>
    require.config({
      map: {
        '*': {
          'ipyvolume': 'ipyvolume_v0.4.5',
        }
      }})
</script>
...
</body>
```
So somehow `ipyvolume.js` is getting imported multiple times, I'm not sure why, and I'm not sure if `three.js` is supposed to be imported in a similar way.

2. There's some commented code in `save_ipyvolumejs()` which suggests an earlier version downloaded `three.js` from the `static` folder. I'm not sure what to do with it, and in particular with the pointer to https://github.com/jovyan/pythreejs/issues/109
```python
def save_ipyvolumejs(target="", devmode=False,
                     version=ipyvolume._version.__version_js__, version3js=__version_threejs__):
    ...
    # TODO: currently not in use, think about this if we want to have this external for embedding,
    # see also https://github.com/jovyan/pythreejs/issues/109
    # three_filename = 'three_v{version}.js'.format(version=__version_threejs__)
    # three_filepath = os.path.join(target, three_filename)
    # threejs = os.path.join(os.path.abspath(ipyvolume.__path__[0]), "static", "three.js")
    # shutil.copy(threejs, three_filepath)

    return pyv_filename#, three_filename
```

3. Unlike all the other resources which are downloaded (`require`, `embed`, `font-awesome`, `ipyvolume`), `three.js` does not appear in the `snippet` variable created in `embed_html()`, which is strange.
```python
        fname_pyv = save_ipyvolumejs(scripts_path, devmode=devmode)
        fname_three = save_threejs(os.path.join(scripts_path))
        fname_require = save_requirejs(os.path.join(scripts_path))
        fname_embed = save_embed_js(os.path.join(scripts_path))
        fname_fontawe = save_font_awesome(os.path.join(scripts_path))

        subsnippet = wembed.embed_snippet(widgets, embed_url=rel_script_path+fname_embed,
                                          requirejs=False, drop_defaults=drop_defaults, state=state)
        if not offline_cors:
            # TODO DIRTY hack, we need to do this cleaner upstream
            subsnippet = subsnippet.replace(' crossorigin="anonymous"', '')

        cors_attribute = 'crossorigin="anonymous"' if offline_cors else ' '
        snippet = """
<link href="{rel_script_path}{fname_fontawe}/css/font-awesome.min.css" rel="stylesheet">    
<script src="{rel_script_path}{fname_require}"{cors} data-main='./{rel_script_path}' ></script>
<script>
    require.config({{
      map: {{
        '*': {{
          'ipyvolume': '{fname_pyv}',
        }}
      }}}})
</script>
{subsnippet}
        """.format(rel_script_path=rel_script_path, fname_fontawe=fname_fontawe, fname_require=fname_require,
                   fname_pyv=os.path.splitext(fname_pyv)[0], 
                   subsnippet=subsnippet, cors=cors_attribute)
```

maartenbreddels:
Thanks for you (first!) contribution.
I just got bitten today by the fontawesome url issue, so good timing! 
I think three.js shouldn't be needed actually. In the notebook version threejs is not bundled, for the embedded version it is, see `js/webpack.config.json`, which does not list three as external for dist/index.js, which is used. So you see it trying to fetch three.js?

mathieu1:
> I just got bitten today by the fontawesome url issue, so good timing!

Actually my fix currently doesn't work as intended, because the css file has moved inside the zip. Also, font-awesome 5 renamed lots of things, so I need to use the font-awesome-4 shims. Moreover, the zip is 20MB large so I'm looking for a more lightweight alternative (especially since currently it's going to download the resources files every time `savefig(..., headless=True)` is called).

Actually, it's not clear to me why we need to use offline mode when headless. Is it because python could have access to internet while headless chrome would not?

> So you see it trying to fetch three.js?

Ok so it used to but for some reason that's not the case anymore. Not sure what happened there. Scratching that from the PR.

maartenbreddels:
Let's keep it to 4.7 for fontawesome. And no need to use offline for
headless.

(from mobile phone)

On 7 Feb 2018 22:42, "Mathieu" <notifications@github.com> wrote:

I just got bitten today by the fontawesome url issue, so good timing!

Actually my fix currently doesn't work as intended, because the css file
has moved inside the zip. Also, font-awesome 5 renamed lots of things, so I
need to use the font-awesome-4 shims. Moreover, the zip is 20MB large so
I'm looking for a more lightweight alternative (especially since currently
it's going to download the resources files every time savefig(...,
headless=True) is called).

Actually, it's not clear to me why we need to use offline mode when
headless. Is it because python could have access to internet while headless
chrome would not?

So you see it trying to fetch three.js?

Ok so it used to but for some reason that's not the case anymore. Not sure
what happened there. Scratching that from the PR.

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/maartenbreddels/ipyvolume/pull/104#issuecomment-363920477>,
or mute the thread

mathieu1:
Sounds good: using working url for font-awesome 4.7.0 and disabled the offline mode when headless. Ready for review.

maartenbreddels:
Ok great!
I'm still thinking that it's not ideal that the headless mode will have to download all the files in the browser. I can imagine that someone wants to make many screenshots (maybe for a movie), what would result in many fetches. It would be better to cache the files somewhere, although that is something the browser maybe should do.
I'll wait till someone has that use case, lets keep it simple first.
Thanks a lot!

Issue #105, opened 2018-02-14
Improvement: Current Figure Handling


TimoFriedri:
Hi,

I was struggeling with the way figure handling is implemented. 

This does not work in Ipyvolume but many people are used to do it like this from e.g. matlab.
```
f1 = ipv.figure()
f2 = ipv.figure()
ipv.figure(f1) # make f1 the current figure which is then used e.g. with ipv.movie
```

To make it work you have to do:
```
f1 = ipv.figure(key='f1')
f2 = ipv.figure(key='f2')
ipv.figure('f1') # make f1 the current figure which is then used e.g. with ipv.movie
```
If you know it is done like this, you will understand the documentation. But it took me ages to figure this out. I had to look at the source code in pylab.py. Maybe I am just biased by my Matlab past.

I could imagine to pass a Figure Object to ipv.figure(), then figure() notices the type and searches for that figure in ipv.current.figures.values.
So both usages would work.

Thanks

Bye

PS: Very nice library

maartenbreddels:
Hi Timo,

thanks, and excellent idea, that indeed is expected behaviour. Do you want to try to make a PR for that? I can help if you want.

cheers,

Maarten

TimoFriedri:
Hi Maarten,

I already thought about it. It is right now just limited by free time on my side.
I'll send an PR the next days.

Greets

PR #106, opened 2018-02-15
Added support for ipv.figure(ipv.Figure) syntax

TimoFriedri:
In reference to this issue: #105 

Downside of all the Figure handling atm is that one has to take care not to accidentally use ipv.Figure instead of ipv.figure in your experiments.

### DIFF
--- a/ipyvolume/pylab.py
+++ b/ipyvolume/pylab.py
@@ -13,6 +13,8 @@ from . import examples
 import warnings
 import PIL.Image
 import traitlets
+import uuid
+
 try:
     from io import BytesIO as StringIO
 except:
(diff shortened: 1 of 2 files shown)


maartenbreddels:
Looking good, thanks!
Would you mind adding a unittest, ala:
https://github.com/QuantStack/ipysheet/blob/master/ipysheet/test_all.py#L10
you can add it to:
https://github.com/maartenbreddels/ipyvolume/blob/master/ipyvolume/test_all.py
and install/run `py.test`

TimoFriedri:
Hi,

I will do it soon. I have some issues with py.test. Complaining about missing traittypes which is definitely installed in mit env.

I have to fix that first

TimoFriedri:
Hi, I need support.

I have install ipyvolume for the fork in an Conda Env via pip install -e.
In this env I have also the package traittypes.
If I import traittypes now, the traittypes.py file in ipyvolume is used instead of the package traittypes, which break a lot, including running the py.test

In my env where I have install ipyvolume from conda-forge, everything works as expected.

Any idea?
Thanks

TimoFriedri:
I have tried to fix my traittypes import issue but haven't succeeded yet.

I run 
 `py.test ./ipyvolume/test_all.py`

Then I get that Error:

``` 
py.test ./ipyvolume/test_all.py 
===================================== test session starts ======================================
platform linux -- Python 3.6.1, pytest-3.0.7, py-1.4.33, pluggy-0.4.0
rootdir: .../github/ipyvolume, inifile:
collected 0 items / 1 errors 

============================================ ERRORS ============================================
____________________________ ERROR collecting ipyvolume/test_all.py ____________________________
ImportError while importing test module '.../github/ipyvolume/ipyvolume/test_all.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
ipyvolume/__init__.py:5: in <module>
    from .widgets import *
ipyvolume/widgets.py:4: in <module>
    from traittypes import Array
E   ModuleNotFoundError: No module named 'traittypes'
!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!
=================================== 1 error in 0.52 seconds ==================================== 
```

For some reason it is using the traittypes.py in ipyvolume. If I open ipython and import traittypes, I can check with `traittypes.__file__` that the correct package is used.

maartenbreddels:
This only happens with py.test?
can you give me the output of:
$ `which python`
and
$ `which py.test`

TimoFriedri:
Thanks, you gave me the correct idea for what might be wrong. pytest was still not installed in my env, only in root. This caused the issue. I have somehow overseen this the whole time.

I'll update the PR soon.

maartenbreddels:
Excellent, thanks!

PR #112, opened 2018-03-02
Add top and bottom to plot_plane

gepcel:
I'm not sure about something. Please review this.


maartenbreddels:
Thanks, there isn't a proper convention for the faces to face inside or outside.. that should be fixed, but I'll merge this first!

PR #114, opened 2018-03-06
FIX: surface color

kingjr:
The example below couldn't run.
```python
import os
import numpy as np
import matplotlib.pyplot as plt
from matplotlib import cm
import ipyvolume.pylab as p3

X = np.arange(-5, 5, 0.25*1)
Y = np.arange(-5, 5, 0.25*1)
X, Y = np.meshgrid(X, Y)
R = np.sqrt(X**2 + Y**2)
Z = np.sin(R)
colormap = cm.coolwarm
znorm = Z - Z.min()
znorm /= znorm.ptp()
znorm.min(), znorm.max()
color = colormap(znorm)
p3.figure()
mesh = p3.plot_surface(X, Z, Y, color=color[...,0:3])
p3.show()
```


maartenbreddels:
Great, thanks!

PR #117, opened 2018-03-12
Add option to define x,y,z limits for isosurfaces

OliverEvans96:
I needed this and didn't see it available, so I added it. Suggestions are welcome!

### DIFF
--- a/ipyvolume/pylab.py
+++ b/ipyvolume/pylab.py
@@ -552,7 +552,7 @@ def transfer_function(level=[0.1, 0.5, 0.9], opacity=[0.01, 0.05, 0.1], level_wi
     return tf
 
 
-def plot_isosurface(data, level=None, color=default_color, wireframe=True, surface=True, controls=True):
+def plot_isosurface(data, level=None, color=default_color, wireframe=True, surface=True, controls=True, extent=None):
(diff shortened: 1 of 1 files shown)


maartenbreddels:
Hi Olvier,

thanks alot, this totally makes sense. I've already done this for [volshow](https://github.com/maartenbreddels/ipyvolume/blob/5f206ecf6a3e688895a2e9fabbc50650f05d7b74/ipyvolume/pylab.py#L605) which is using the extent argument, similar to matplotlib's imshow. Would you mind changing it to be similar to that?

cheers,

Maarten

PS: sorry for getting so late at this.

OliverEvans96:
How's that? Short and sweet. It's fascinating how it often takes a long time to arrive at simple solutions. Sorry for the delay!

Oliver

OliverEvans96:
As a note (perhaps to myself), similar to what you've done for `volshow`, this modifies the extent of the data, not the extent of the figure. However, if the data exceeds the figure, it will grow, but it will not shrink when the figure extent is larger than the data. For specifying figure extens, `ipv.xlim`, `ipv.ylim` and `ipv.zlim` can be used as in matplotlib.

maartenbreddels:
looking good, and I believe those statements are correct.
Thanks Oliver!

PR #121, opened 2018-03-20
Bug fix: Updated font awesome zip file location

DougRzz:
The ipyvolume.embed.embed_html("example.html", widgetName , offline=True, devmode=False) command had an error. Could not find the font-awesome zip file.

This commit redirects download from https://use.fontawesome.com/releases/v5.0.8/fontawesome-free-5.0.8.zip


maartenbreddels:
Hi Doug,

this was already fixed in https://github.com/maartenbreddels/ipyvolume/pull/104 except the version was not changed.

cheers,

Maarten

Issue #61, opened 2017-09-05
Scatterplots w/ sprites


choldgraf:
Is it possible to change scatterplot markers to that they "look" like they're a 2D point (e.g., a flat plane that always faces the camera no matter the orientation).

E.g., this is the default behavior in matlab, so I think a lot of people will have this in mind when they look at the scatter function here:

https://www.mathworks.com/help/matlab/ref/scatter3.html?requestedDomain=www.mathworks.com

maartenbreddels:
Yes, it's been on my TODO list for quite a while, it's called sprites, thanks for opening this. It should also give a better performance for many points, say in the order of 10^6.

dakami:
Please also add opacity support; when plotting large numbers of points being able to alter the blend factors is huge.  (Came here to file that bug, actually.)

empeeu:
I'd be happy to try implementing this if  you can give me a few hints of where to start.

maartenbreddels:
This is implemented, see referred PR's

PR #122, opened 2018-03-20
WIP: FEATURE: Sprite rendering in scatter view

mpu-creare:
Goal is to supply 'markers' or 'geos' that are rendered as sprites. This will greatly enhance speed of rendering for point clouds with > 10^6 points. 

This is to address issue #61

### DIFF
--- a/ipyvolume/test_all.py
+++ b/ipyvolume/test_all.py
@@ -110,6 +110,24 @@ def test_figure():
         for debug in [True, False]:
             p3.figure(debug=debug, controls=controls)
 
+def test_context():
+    f1 = ipv.figure(1)
+    f2 = ipv.figure(2)
+    f3 = ipv.figure(2)
+
+    assert ipv.gcf() is f3
+    with f2:
+        assert ipv.gcf() is f2
+    assert ipv.gcf() is f3
+    # test nested
+    with f2:
+        assert ipv.gcf() is f2
+        with f1:
+            assert ipv.gcf() is f1
+        assert ipv.gcf() is f2
+    assert ipv.gcf() is f3
+
+
 def test_limits():
     f = p3.figure()
     p3.xlim(-10, 11)
--- a/ipyvolume/widgets.py
+++ b/ipyvolume/widgets.py
@@ -161,6 +161,27 @@ class Figure(ipywebrtc.MediaStream):
         self._selection_handlers = widgets.CallbackDispatcher()
         self.on_msg(self._handle_custom_msg)
 
+    def __enter__(self):
+        """Sets this figure as the current in the pylab API
+
+        Example:
+        >>> f1 = ipv.figure(1)
+        >>> f2 = ipv.figure(2)
+        >>> with f1:
+        >>>  ipv.scatter(x, y, z)
+        >>> assert ipv.gcf() is f2
+        """
+
+
+        import ipyvolume as ipv
+        self._previous_figure = ipv.gcf()
+        ipv.figure(self)
+
+    def __exit__(self, type, value, traceback):
+        import ipyvolume as ipv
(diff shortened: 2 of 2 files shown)


mpu-creare:
There are a few problems with this implementation, I'll need some feedback. 
* I couldn't find a way to render points as sprites using `THREE.Mesh` so I switched to `THREE.Points` and added a new material
* The scaling of the points are off -- it seems like everything is scaled to a 1x1 box
    * Is that true?
    * Where does that scaling happen? 
    * Is it possible to turn that off... similar to `axis('equal')`
    * Right now, if I render an ellipse of points it will look like a sphere

Here's a screenshot of the second issue... 
image

mpu-creare:
A few more points: 
* Major changes are in `scatter.js`
* I added and `examples` folder with two tests plots (I'll strip these out when the PR is finished)
    * These were created with `embed_html(..., offline=True)`
    * I manually edited the .html to point to `../js/dist/index.js` which was created by the `npm install`

maartenbreddels:
Hi,

Thanks, this was on my wish list for quite a long time!
Yes, there is some custom transformation going on, and I think this requires a custom shader. I'm not sure the Points can have a custom shader, it should behave similar as the mesh and scatter shaders (doing animation/interpolation) and on top of that the transformation.
I'm quite busy at the moment, but I'm gonna try my best to support you, hope I can give you some more pointers soon.

cheers,

Maarten

mpu-creare:
I spent most of the day trying to track down that transformation and I stumbled across that custom shader -- if you can give me a hint where that happens, I'm still interested.
*  I thought it happens in the vertex shader https://github.com/maartenbreddels/ipyvolume/blob/master/js/glsl/scatter-vertex.glsl#L82
* BUT, as far as I can tell `origin` and `size_viewport`  https://github.com/maartenbreddels/ipyvolume/blob/master/js/glsl/scatter-vertex.glsl#L51
are [0, 0, 0] and [1,1,1] respectively 
   * based on `last_scatter.material.uniforms` from the browser console

I see the https://github.com/maartenbreddels/ipyvolume/blob/master/js/glsl/scatter-fragment.glsl#L26 SPRITE block all ready for me. I will re-implement this as a custom shader then and revert all the code I've written sofar... so stay tuned for that.

mpu-creare:
Oh, and thanks for the reply Maarten. I can appreciate the business. 

I very much appreciate `ipyvolume`. I played around with `pythreejs` for a bit before finding your project -- which suits my use-cases much better. I don't care much for jupyter notebook so the embedded html is amazing.

maartenbreddels:
Hi, you closed the PR, you don't want to continue?

empeeu:
No, I'm working of a shader version of this -- nearly done. Github apparently closed this for me after I invoked some serious git-fu. 

Stay tuned, I'm nearly there -- and it's much cleaner with just the shader.

maartenbreddels:
Yes, I planned to work on sprites a while ago.. so there is some leftovers for this. I'm actually in need for this really soon, so I'm eager to test is as well.
origin is basically the coordinate of 1 of the corners of the box, so https://github.com/maartenbreddels/ipyvolume/blob/master/js/glsl/scatter-vertex.glsl#L82 scales it first to the range [0, 0, 0], [1, 1, 1] and by subtracting -0.5 to [-0.5, -0.5, -0.5], [0.5, 0.5, 0.5]

larsoner:
I haven't looked at the code, but if you are using `GL_POINT_SPRITE` for markers in scatter plots, VisPy has a bunch of styles you can use / adapt if you want:

https://github.com/vispy/vispy/tree/master/vispy/glsl/markers

maartenbreddels:
I thought of using textures only, but thanks for the tip!

Issue #97, opened 2018-01-10
Multiframe volume data, non-cubical data domain and why png base64 and not binary array?


blauwepiet:
I was looking at implementing non-cubical data domains, which kinda worked out with a lot of hacking because it is all hard-coded to be cubical but now all meshes and the domain is scaled based on the limits of the x,y,z domain bounds of the dataset.

But now I was looking at also visualizing volumes together with the meshes but I stumbled on an error while creating the png for the volume (32x32x48). The following error:
```
...
 /home/casperl/Projects/HemoSphere/src/ipyvolume/ipyvolume/serialize.pyc in cube_to_png(grid, vmin, vmax, file)
     81                 warnings.simplefilter("ignore")
     82                 img = PIL.Image.frombuffer("RGBA", (image_width, image_height), data, 'raw')
---> 83                 img.save(file, "png")
     84         return (image_width, image_height), (grid.shape[2], grid.shape[1]), rows, columns, grid.shape[0]
     85 

/home/casperl/anaconda2/envs/hemosphere2/lib/python2.7/site-packages/PIL/Image.pyc in save(self, fp, format, **params)
   1926 
   1927         try:
-> 1928             save_handler(self, fp, filename)
   1929         finally:
   1930             # do what we can to clean up

/home/casperl/anaconda2/envs/hemosphere2/lib/python2.7/site-packages/PIL/PngImagePlugin.pyc in _save(im, fp, filename, chunk, check)
    795 
    796     ImageFile._save(im, _idat(fp, chunk),
--> 797                     [("zip", (0, 0)+im.size, 0, rawmode)])
    798 
    799     chunk(fp, b"IEND", b"")

/home/casperl/anaconda2/envs/hemosphere2/lib/python2.7/site-packages/PIL/ImageFile.pyc in _save(im, fp, tile, bufsize)
    482             if o > 0:
    483                 fp.seek(o, 0)
--> 484             e.setimage(im.im, b)
    485             if e.pushes_fd:
    486                 e.setfd(fp)

SystemError: tile cannot extend outside image
```
And then I also notice that the image width is restricted to 2048, why is that? Why not 4096x4096? This is common max tex size as far as I know. And why is the volume send to the client with png base64 instead of the binary array as the rest of the data? 

Also was wondering why the volume is integrated in the figure widget and not a stand-alone widget? Is that because of legacy coding choices? Because it would look more need if it is also a separate widget like the meshes.

[...]

PR #123, opened 2018-03-22
FEATURE: Scatter sprite rendering

mpu-creare:
Goal is to supply 'markers' or 'geos' that are rendered as sprites. This will greatly enhance speed of rendering for point clouds with > 10^6 points.

This is to address issue #61

This continues where https://github.com/maartenbreddels/ipyvolume/pull/122 left off. 

Basically added a new section for the shader, and added two `geos` that are just planes. 
Not sure if this is the `best` way to do this, but the x-y-z positions are projected into clip space, then the `positions` from the `geo` are used to offset those points -- in clipspace.


js/src/scatter.js
```diff
@@ -78,6 +95,13 @@ var ScatterView = widgets.WidgetView.extend( {
             visible: this.model.get("visible") && this.model.get("visible_markers")
             })
 
+        this.material_sprite = new THREE.RawShaderMaterial({
```
mpu-creare: New material -- basically add a #define USE_SPRITE

js/src/scatter.js
```diff
@@ -263,10 +287,15 @@ var ScatterView = widgets.WidgetView.extend( {
         // add atrributes to the geometry, this makes the available to the shader
         current.add_attributes(instanced_geo)
         previous.add_attributes(instanced_geo, '_previous')
-
-	    this.mesh = new THREE.Mesh(instanced_geo, this.material );
+        var material;
```
mpu-creare: This section below here swaps out the new material. I wasn't sure what material_rgb does, or where it gets used... Not sure if this is the best way to do this swap.

js/glsl/scatter-vertex.glsl
```diff
@@ -50,7 +50,31 @@ attribute vec3 color_previous;
 void main(void) {
     vec3 origin = vec3(xlim.x, ylim.x, zlim.x);
     vec3 size_viewport = vec3(xlim.y, ylim.y, zlim.y) - origin;
+#ifdef USE_SPRITE
+    vec3 vector = v;
```
maartenbreddels: vectors are only used for the direction of a 3d glyph/geo, to make quiver plots possible, we can ignore them

js/glsl/scatter-vertex.glsl
```diff
@@ -50,7 +50,31 @@ attribute vec3 color_previous;
...
+                vec4(pos,1.0);
+    // Test vector used to correct scale in clipspace
+    vec4 posxy_test = projectionMatrix *
+                modelViewMatrix *
+                vec4(1.0,0.0,0.0,1.0);
+    float posxy_mag = sqrt(posxy_test.x*posxy_test.x +
+                           posxy_test.y*posxy_test.y +
+                           posxy_test.z*posxy_test.z);   
+    s = s * posxy_mag; 
+    gl_Position = posxy + vec4((position.x-0.5)*s, (position.y-0.5)*s, 0, 0);
```
maartenbreddels: position is not set, only for the line, and I think this code can probably share a lot with the #ifndef AS_LINE block.

maartenbreddels: gl_Position is set again on line 109 as is vertex_position and vertex_uv.

js/src/scatter.js
```diff
@@ -50,7 +64,10 @@ var ScatterView = widgets.WidgetView.extend( {
             arrow: this.geo_arrow,
             sphere: this.geo_sphere,
             cat: this.geo_cat,
+            point: this.geo_point,
+            tri: this.geo_tri
         }
+        this.sprite_geos = ['point', 'tri'];
```
maartenbreddels: what about prefixing the names?
sprite_point or 2d_point

js/src/scatter.js
maartenbreddels: please use sprite_material to be consistent with the rest of the naming (or change chose ;-) )

js/src/scatter.js
maartenbreddels: material_rgb should write rgb color corresponding to the normalised (0, 0, 0) (1, 1, 1) which is used for the volume rendering to know where opaque objects are, so where to stop the ray.

js/glsl/scatter-vertex.glsl
```diff
@@ -50,7 +50,31 @@ attribute vec3 color_previous;
...
+    vec3 pos = (mix(position_offset_previous, position_offset, vec3(animation_time_x, animation_time_y, animation_time_z))
+                - origin) / size_viewport - 0.5;
+    vec4 posxy = projectionMatrix *
+                modelViewMatrix *
+                vec4(pos,1.0);
+    // Test vector used to correct scale in clipspace
+    vec4 posxy_test = projectionMatrix *
+                modelViewMatrix *
+                vec4(1.0,0.0,0.0,1.0);
+    float posxy_mag = sqrt(posxy_test.x*posxy_test.x +
```
maartenbreddels: what does this do? I noticed when rotating example/0001.html that the size of the sprite changes, what is the idea behind it?

mpu-creare: Good question. At the moment `size` is a percentage of the viewport -- which I found very useful. Without this code, `size` refers to a number of pixels... but the size of the sprites shouldn't change, which means I worked in size attenuation unintentionally... I'll rethink this. 

Unless you're happy with `size` meaning a number of pixels in the case of sprites.

js/glsl/scatter-vertex.glsl
maartenbreddels: ignore this, it is begin used!

js/glsl/scatter-vertex.glsl
```diff
@@ -50,7 +50,31 @@ attribute vec3 color_previous;
...
+                           posxy_test.y*posxy_test.y +
+                           posxy_test.z*posxy_test.z);
```
maartenbreddels: `float posxy_mag = length(posxy_test.xyz);` should work

js/glsl/scatter-vertex.glsl
maartenbreddels: no, it would be nice if it is consistent. It seems right actually.

js/glsl/scatter-vertex.glsl
empeeu: Again, nested ifdefs... that code doesn't get compiled when `USE_SPRITE` is defined. I'll play around and see if I can get the #elif working -- I don't write a lot of shaders. :)

js/src/scatter.js
empeeu: I like 2d_point -- will change this.

js/glsl/scatter-vertex.glsl
empeeu: Also, I considered sharing a lot of the code with `AS_LINE` block, but there was enough different code, and this ended up being considerably more readable.

js/src/scatter.js
maartenbreddels: Yes, it is when volume rendering and scatters are mixed in one plot, for instance:
```python
fig = ipv.figure()
cube = ipv.datasets.aquariusA2.fetch().data
ipv.volshow(cube, extent=[[-4, 4]]*3)
g = ipv.examples.gaussian(N=100);
```
Maybe crank up the opacity for the first level to see the scatter points fade away. If you now change the geo to a sprint, it will not render properly.
For debugging, you can set: `fig.show = 'Back' # or 'Front' or 'Volume'`

ipyvolume/pylab.py
```diff
@@ -37,7 +37,7 @@ def _docsubst(f):
     "color2d"] = "color for each point/vertex string format, examples for red:'red', '#f00', '#ff0000' or 'rgb(1,0,0), or rgb array of shape (2, N, 3) or (S, 2, N, 3)"
 _doc_snippets[
     "size"] = "float representing the size of the glyph in percentage of the viewport, where 100 is the full size of the viewport"
-_doc_snippets["marker"] = "name of the marker, options are: 'arrow', 'box', 'diamond', 'sphere'"
+_doc_snippets["marker"] = "name of the marker, options are: 'arrow', 'box', 'diamond', 'sphere', 'point_2d', 'tri_2d'"
```
maartenbreddels: Yes, good choice, if we'd prefix it, we could never turn them into valid Python variable names, since we might consider doing an Enum in the future.

maartenbreddels:
I've made some simplifications to the shader, do you mind if I push these changes?

empeeu:
Yes, please go ahead and push the simplifications. 

Also, when this is ready, let me know and I'll strip out the examples -- no need for those to be in the repo.

maartenbreddels:
I can't push to your branch, did you disable it?

maartenbreddels:
Also want to let you know I'm working on getting textures in.

mpu-creare:
It might be a permissions issue. I just gave  you access to my fork -- let me know if that works. 

Re: Texture -- excellent! That was on my list, but please feel free if you have the bandwidth. 

Changes in the frag shader should be pretty straightforward... 

For textures, here are a few thoughts, take 'em or leave em:
* For a user-specified custom texture
    * `marker = 'sprite_2d'`
    * user supplies `marker_data` (image for texture)
        * This field could also potentially be used for custom glyphs
    * add texture to `sprite_material`
* to render the texture:
    * The `geo_point_2d` could be used as the template, and a square maps nicely to a square, but uses 2 faces
    * Alternatively you could use a right-angle triangle -- this is only 1 face, not sure if that's more efficient, but you need some logic to deal with the parts of the triangle that do not map to the square
image

maartenbreddels:
It was an issue on my side because of 2FA.
I've added a texture similar to what I've done in mesh, it's a bit of a code duplication, but I'll see if I can refactor that later.
Some other changes:
 * point is a small square
 * square is actually 2x2.. and i rescale these coordinates to uv (texture) coordinates. I think all the geometries should be 1x1 in size, so that is really is a percentage of the viewport.
I'll push the change for the textures next, so you can see if it suits your needs.

js/glsl/scatter-vertex.glsl
```diff
@@ -50,50 +50,55 @@ attribute vec3 color_previous;
...
-        + (mix(position_offset_previous, position_offset, vec3(animation_time_x, animation_time_y, animation_time_z))
-                - origin) / size_viewport - 0.5;
+    vec3 model_pos = (mix(position_offset_previous, position_offset, animation_time) - origin) / size_viewport - 0.5;
     //vec3 pos = (pos_object ) / size;// - 0.5;
-#else
-    vec3 pos = (mix(position_previous, position, vec3(animation_time_x, animation_time_y, animation_time_z))
-                - origin) / size_viewport - 0.5;
+    #ifdef USE_SPRITE
+        // if we are a sprite, we add the position in view coordinates, and need to 
+        vec4 view_pos = modelViewMatrix * vec4(model_pos, 1.0);
```
mpu-creare: USE_SPRITE doesn't need move_to_vector -- you can save some computation in that case (perhaps the compiler takes care of this already...)

mpu-creare: That was a nice refactor on the vertex shader.

maartenbreddels: Actually, we could use this to rotate the sprites, what do you think? The z component could be used for maybe a stretch factor?

mpu-creare: That sounds like a great idea -- particularly for non-square textures...

maartenbreddels:
I don't have any more new things, so if you want to polish it up, maybe remove the demos, update the docstring, I hope I can merge this after the weekend!

mpu-creare:
I added transparency for the textures -- although it looks like the draw order is a little wonky. 


I'll go ahead and clean this up -- certainly by the end of the weekend.

maartenbreddels:
Great, I'm was also playing with that
Try adding this to the material:
```
            depthWrite: false
```
What I use now (for almost black textures without alpha):
```
            blending: THREE.CustomBlending,
            blendSrc: THREE.SrcColorFactor,
            blendDst: THREE.OneFactor,
            blendEquation: THREE.AddEquation,
            transparent: true,
            alphaTest: 0.01,
            depthWrite: false
```
But it's never gonna be perfect, for that you need to sort by depth, or wait for @blauwepiet his depth peeling (https://github.com/maartenbreddels/ipyvolume/issues/97)
We could expose the attributes btw, for some cases it's good enough, but it's maybe also a good reason to start depending on pythreejs, and expose the whole material.

mpu-creare:
`depthWrite: false` works quite well. kitty cloud for the win!



What do you think the behavior should be if an array of colors is also defined? Should the colors mix with the texture? only use the texture alpha channel? only use the texture intensity? Any comments? 

+1 for the pythreejs dependency. No reason why both projects should wrap three.js in python... though you are doing a lot of good custom three.js development on the javascript side. 

ipyvolume is much better for scientific visualizations, and complements pythreejs's goals.

maartenbreddels:
Oops, my mistake, that should have been:
```javascript
this.triangle_2d = new THREE.CircleGeometry(1, 3, Math.PI/2);
```
Maybe rename tri_2d to triangle_2d?

maartenbreddels:
I think we shouldn't set `depthWrite: false`, lets keep this PR as clean as possible. I already experimented a bit with using a pythreejs material, than we can change all properties!

mpu-creare:
I got myself into a little bit of trouble with `git` as I was stripping out the example files so I ended up having to squash some commits -- as a result, your changes aren't reflected in the history. 

Other than that, I think this one is relatively clean at this point.

maartenbreddels:
Thanks a lot for the work, an awesome first time contribution! 👍

maartenbreddels:
On master you can now use the materials, make sure you run `$(cd js; npm install)` to make sure threejs is up to date and try sth like this:
```python
mw.material.blending = pythreejs.BlendingMode.CustomBlending
mw.material.blendSrc = pythreejs.BlendFactors.SrcColorFactor
mw.material.blendDst = pythreejs.BlendFactors.OneFactor # or 'OneFactor'
mw.material.blendEquation = 'AddEquation'
mw.material.transparent = True
mw.material.depthWrite = False
mw.material.alphaTest = 0.1
```
where mw is a Scatter or Mesh.

PR #125, opened 2018-03-28
Fix plot_isosurface docstring

jackieleng:
Some small fixes.

### DIFF
--- a/docs/source/mesh.ipynb
+++ b/docs/source/mesh.ipynb
@@ -111,8 +111,8 @@
     "Y = V\n",
     "Z = X*Y**2\n",
     "p3.figure()\n",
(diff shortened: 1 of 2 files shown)


maartenbreddels:
Hi Jackie,

thanks a lot, the `X, Z, Y` was deliberate since Y is up.. but that actually touched on an issue that for many cases you want Z to be up.

cheers,

Maarten

jackieleng:
Okay, I've reverted the changes to mesh.ipynb. I was indeed thinking Z should be up/down, but I'm not too familiar with the semantics of the examples.

maartenbreddels:
No problem, thanks you for your contribution!

Issue #138, opened 2018-05-25
Problems with alpha blending and renderer properties


OlehKSS:
I am trying to implement alpha blending for two meshes. I am using material property of the mesh class and try to implement custom blending. Color blending seems to work just fine, but if I switch to alpha instead of blending of two meshes I see only the top mesh. Is it possible to implement alpha blending in ipyvolume in this way?

For the [WebGL Renderer](https://threejs.org/docs/#api/renderers/WebGLRenderer) they have an option for enabling transparency buffer, by default it is disabled. Can it be the reason why I don't receive correct alpha? Where can I see or change ipyvolume renderer instance properties?

maartenbreddels:
You could be right, and that would be [here in FigureModel](https://github.com/maartenbreddels/ipyvolume/blob/master/js/src/figure.js#L487).
I think it's good to have this on by default, if you make a PR I'm happy to accept it. Later on we could have it as an option, but I'm ok with this solution for now.

OlehKSS:
Three.js doesn't support alpha channel in colors according to this [pull request](https://github.com/mrdoob/three.js/pull/6898) . So, I will need to try a workaround with `vector4` and a custom shader, as in the [issue](mrdoob/three.js#2730). Could you suggest where I can start to dig in ipyvolume code in order to do that?

maartenbreddels:
Actually, the color array can of shape (M, N, 3 or 4) or (N,3 or 4), where 4 is for rgba values. To be honest, I can remember this not working, please try it out and report back, if you want try out yourself. 
 * https://github.com/maartenbreddels/ipyvolume/blob/d43092780ba25cad428dff784accb03c1964edfd/ipyvolume/widgets.py#L50 this is where color is defined to be serialized using the color_serialization
 * That refers to https://github.com/maartenbreddels/ipyvolume/blob/d43092780ba25cad428dff784accb03c1964edfd/ipyvolume/serialize.py#L271 which is using https://github.com/maartenbreddels/ipyvolume/blob/d43092780ba25cad428dff784accb03c1964edfd/ipyvolume/serialize.py#L245 to convert it to something that goes over the wire.
 * At https://github.com/maartenbreddels/ipyvolume/blob/d43092780ba25cad428dff784accb03c1964edfd/js/src/mesh.js#L410 you can find that on the front-end we use the color_or_json to convert if from the wire format to something JS friendly
[...]

PR #139, opened 2018-05-30
Solution for #138

OlehKSS:
This is a draft solution of #138.
I have several questions:

1. During editing of the JavaScript code, I saw that you were using ES5, is it because of compatibility with older browsers (IE)? Can I use ES6 in code (like classes, lambdas, etc.)? As far as I know browser support should be almost full, [Compatibility table](http://kangax.github.io/compat-table/es6/).

2. I've been editing values.js module in order to add support for alpha channel. Should I keep it like that or should I create another class, that will work with vector4, instead of vector3 colors? I also thought to rewrite this class using ES6 class syntax.

3. For the colors that are provided as array, but do not have alpha channel I'll need to append it as 1.0, so far I did on the Python side, but I think it can be done on JavaScript side as well. As far as I understand 
serializing/deserialing doesn't affect alpha channel in any way, since it creates 1-D array before transferring data  to the front-end side.

4. I have deleted some [unreachable code](https://github.com/maartenbreddels/ipyvolume/compare/master...OlehKSS:olehkss/138#diff-0dc8f08cfeae91aad971c8f2fa220caaL192) and commented code, [example1](https://github.com/maartenbreddels/ipyvolume/compare/master...OlehKSS:olehkss/138#diff-764120e894f671fd485950ffd11abd64L267) and [example2](https://github.com/maartenbreddels/ipyvolume/compare/master...OlehKSS:olehkss/138#diff-2399ea0769b50adf87884f1c1a8ea493L357). Is it safe to delete? Did you plan to use it somehow?


OlehKSS:
*Example of alpha blending*
ipyvolume 9

White brain mesh with alpha = 0  with blue regions (alpha = 0.3) on to of the grey brain mesh.
Custom blending parameters used for the white mesh material:
```mat.blending = BlendingMode.CustomBlending
mat.blendDst = BlendFactors.OneMinusSrcAlphaFactor
mat.blendEquation = Equations.AddEquation
mat.blendSrc = BlendFactors.SrcAlphaFactor

mat.alphaTest = 0.1
mat.transparent = True
mat.side = Side.DoubleSide
```

There is some white glowing on the edges of blue regions and throughout the mesh, I don't know what can be the cause of those, do you have any idea? I want to obtain something as with color blending.

*Color Blending Example*
ipyvolume 4

larsoner:
That doesn't look like glow to me it looks like the lighter curvature color showing up? Try setting the curvature color to a uniform gray instead of binary light/dark and see if the"glow" goes away. 

Also I think the curvature is backward, gyri should be light and sulci dark. 

But it looks like it's working!

As for es5 vs 6 with GL (I assume that's what we're talking about?) IIUC it's best to stick with the lowest version reasonably possible in order to maintain highest device and application compatibility. How hard would it be to stick with 5?

larsoner:
Also can you add to the todo list (after alpha blending) to get surface normals working? It will fix / smooth out the appearance of the individual triangles

larsoner:
Oh I was looking at the color blending example, I see what you mean about the glow in the alpha example now. No immediate ideas but I'll think about it.

larsoner:
@OlehKSS for me it looks fine:

screen shot 2018-05-30 at 21 49 58

See https://github.com/OlehKSS/mne-gsoc2018-3d/pull/2. It looks like the same changes you list above, not sure why it's not working for you?

OlehKSS:
@larsoner Sticking to es5 will not require anything additionally, but with es6 version of JavaScript code will look nicer in general and more handy features will be available.

OlehKSS:
@maartenbreddels I have updated pull request and removed most of the styling changes.

Could you explain more on colors as arrays on strings? Because I didn't see that part in documentation, so I don't fully understand what these array will look like and which purpose they are going to serve? As I explained before, I will need to add an alpha channel to the colors specified without one, for now I've done it on Python side. However, I think it might be easier to handle all the cases on JS side, after deserialization, I am not sure where to put this code.

Concerning JavaScript version, as far as I understand you are using Webpack as your module manager and builder, I suppose it can be reconfigured to change ES6 code into ES5, I am not sure how easy it will work with TypeScript. Nevertheless, I suppose it should be done in a separate pull request, alongside code style changes that I've done before.

choldgraf:
Also a note that travis isn't happy:

https://travis-ci.org/maartenbreddels/ipyvolume/jobs/387141398#L973

Making travis happy could be a way to determine how trivial it'd be to support newer versions of ES

OlehKSS:
@maartenbreddels 
I have a problem with Travis, with is caused by the fact that on Python side we have pythreejs version 0.85 and on JavaScript side we have threejs version 0.91. Have been facing this kind of problems before?

js/glsl/mesh-fragment.glsl
```diff
@@ -11,10 +11,10 @@ uniform float animation_time_texture;
 
 void main(void) {
 #ifdef USE_RGB
-    gl_FragColor = vec4(vertex_color.rgb, 1);
+    gl_FragColor = vec4(vertex_color.rgb, 1.0);
 #else
 #ifdef AS_LINE
-    gl_FragColor = vec4( vertex_color.rgb, 1.0 );
+    gl_FragColor = vec4(vertex_color.rgb, 1.0);
```
maartenbreddels: Why not vertex_color.a as well?

ipyvolume/widgets.py
```diff
@@ -64,6 +64,28 @@ def _default_material(self):
     def _default_line_material(self):
         return pythreejs.ShaderMaterial()
 
+    @traitlets.observe('color')
+    def _observe_color(self, change):
+        '''Checks dimentions of the array of colors and adds alpha channel if needed.'''
+        new_color = change['new']
+
+        if type(new_color) == np.ndarray:
```
maartenbreddels: better use `isinstance(new_color, np.ndarray)` to support subclasses.

ipyvolume/widgets.py
```diff
@@ -64,6 +64,28 @@ def _default_material(self):
...
+                    color_alpha = np.ones((rows, 4))
+                    color_alpha[:, :-1] = new_color
+                    self.color = color_alpha
+
+            elif (len(new_color.shape) == 1) and (new_color.shape[0] == 3):
+                # we have one row, without alpha channel, we'll add it here
+                color_alpha = np.ones(4)
+                color_alpha[:-1] = new_color
+                self.color = color_alpha
+
```
maartenbreddels: I'm not sure about this one, I have a preference for not changing the trait on the Python side, but only at the moment it goes into the Attribute at the js side, or as an intermediate solution, do this transformation in serialize.py::color_to_binary_or_json.
The reason for this is that eventually, we want to have things on the js side linked to other libraries, that may not support alpha.

maartenbreddels:
If you rebase against master, you should get a green light again!

choldgraf:
woot! nice job @OlehKSS and thanks for helping out @maartenbreddels !

maartenbreddels:
Could you maybe share some notebooks and data to reproduce the plots above? Maybe it's a good idea to put them in the docs, and similar to the other datasets/examples (e.g. see ipyvolume/examples.py and datasets.py) provide an easy way to download them.

choldgraf:
we'll get there eventually @maartenbreddels :-)

maartenbreddels:
It looks so cool, I want it now :)

(from mobile phone)

PR #142, opened 2018-06-11
Values class styling fixes and TypeScript. Other styling fixes.

OlehKSS:
Styling fixes made while solving #138. Values class was rewritten using TypeScript.


maartenbreddels:
Great!
The typescript was unexpected though :)
What do you think, should we move to typescript completely?

OlehKSS:
Oh, I saw TypeScript as dependency in the package.json and I thought that you was planning to use it. So, do you plan to use it? 

I have been working with TypeScript for some time, it is easier to work with than with regular JavaScript. Moreover, there are not many differences between TypeScript and ES6, since TypeScript works just as superset of JavaScript with static type checking. Another good thing, there is no need to do move all code base at once, it can be done gradually, file by file.

OlehKSS:
@maartenbreddels I have added a small change to WebGL configurations for working with alpha blending. Previously, I was obtaining undesirable white glow around borders of transparent/opaque regions for some cases and the introduced change fixes it. Should I open a separate pull request for this or can it go along styling changes?
Here are several resources that I used for information about alpha blending configurations in WebGL:
[WebGL and Alpha](https://webglfundamentals.org/webgl/lessons/webgl-and-alpha.html)
[WEBGL TRANSPARENCY AND ALPHA BLENDING PROBLEM](http://in2gpu.com/2014/04/11/webgl-transparency/)

maartenbreddels:
Looking good, I'm on board on the typescript, and that brings me to something else. How do you feel about having unittests for the JS part. I've set this up for [ipysheet](https://github.com/QuantStack/ipysheet/) and it helped me tremendously in speeding up dev. I've never had the time to set this up, buy maybe you value this as well.

About the rendering artifact, that would be better to have in a different PR. To speed up those small PR's you could consider using the [hub tool](https://hub.github.com/)

OlehKSS:
Good, I will create a separate pull request for fixing rendering artifacts. Unittesting is a good idea, I will need to check how you have configured it for ipysheet first and see how much is required to be done.

maartenbreddels:
Happy to talk you through it a bit, like what the workflow is etc.

OlehKSS:
I have opened a separate pull request with changes to WebGL configurations, #145. This pull request contains only TypeScript and styling changes.

OlehKSS:
@maartenbreddels I have added a [pull request](https://github.com/maartenbreddels/ipyvolume/pull/146) with unit test configuration for JavaScript/TypeScript source code.

maartenbreddels:
Looking good, merging!

PR #145, opened 2018-06-16
Changing webgl configurations so it behaves like OpenGL.

OlehKSS:
I have added a small change to WebGL configurations for working with alpha blending. Previously, I was obtaining undesirable white glow around borders of transparent/opaque regions for some cases and the introduced change makes this transition smooth.

*Without configuration changes, white glowing*
ipyvolume 16

*With changes and without white glowing*
ipyvolume 19

### DIFF
--- a/js/src/figure.js
+++ b/js/src/figure.js
@@ -484,7 +484,7 @@ var FigureView = widgets.DOMWidgetView.extend( {
         this.canvas_overlay_container.appendChild(this.canvas_overlay);
         this.canvas_container.appendChild(this.canvas_overlay_container);
 
-        this.renderer = new THREE.WebGLRenderer({alpha: true, antialias: true});
+        this.renderer = new THREE.WebGLRenderer({alpha: false, antialias: true, premultipliedAlpha: false});
 
         this.canvas_renderer_container = document.createElement("div")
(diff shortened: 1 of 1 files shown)


maartenbreddels:
I think it makes sense to have this. Thanks.

maartenbreddels:
I think I have to revert this change. I recently checked out blending, to get it all right, and what I think was going on, is that you have used a multiplied colors (meaning that the rgb colors are already multiplied with alpha), and used for blending `blendSrc: SrcAlphaFactor` insteaf of `OneFactor`, which makes the WebGL pileline multilpy it again. You can indeed circumvent this by saying the whole canvas is not premuliplied, but it is not a good general solution. A better solution is to always (well, mostly) use pre multiplied colors + `OneFactor`.
See https://limnu.com/webgl-blending-youre-probably-wrong/ or https://github.com/mrdoob/three.js/issues/1864#issuecomment-411104710

OlehKSS:
Hello, thanks for the update. I wanted to use `blendSrc: OneFactor` instead of `blendSrc: SrcAlphaFactor`, but I had incorrect color blending with that option. I have found just now that the problem was due to the fact that I have not premultiplied rgb values while calculating a fragment color. I have prepared a small PR #164 for that with the revert of webgl config to the old one.

PR #150, opened 2018-07-04
Fix documentation (build has failed for last 6 months)

OliverEvans96:
Hi Maarten,

I was working on documentation for another project and came to seek some inspiration,
and found that the docs have been failing for a while!

The last successful build was on [2017/10/27](https://readthedocs.org/projects/ipyvolume/builds/6187521/), commit a24d5ec.
The first failing build was on [11/30/2017](https://readthedocs.org/projects/ipyvolume/builds/6343895/), commit d5a620d.

[Here](https://github.com/maartenbreddels/ipyvolume/compare/a24d5ece55c4b36ebd962fe062daecc80dbece99...d5a620dcb35bea8e0f917b2256657e5c4ff2af6a) is the diff between the two versions.
The main thing that changed was updating the version info.
Sphinx was expecting the old version info, and it just needed to be told the new variable name.

I also noticed a few other issues which gave errors on my machine, but don't seem to have changed in that time period (although perhaps they have between a24d5ec and master).
Namely,
- Missing an underline in a section heading
- Nonexistent static directory

The docs build successfully on my laptop after these changes; hopefully they will build on RTD as well!

Cheers,
Oliver


maartenbreddels:
Thanks for looking into it, I'm not sure why you also edited the _static part?

OliverEvans96:
Sphinx complained about not being able to find the directory on my machine; I can't remember whether it was an error or a warning. [Here](https://github.com/QuantStack/xtensor/tree/master/docs/source/_static) is an example of a repo (xtensor) which has a custom stylesheet for their docs, which requires that they have that line in `conf.py`. Since I saw no such thing in ipyvolume, I removed it.

PR #158, opened 2018-07-25
Fixes #156, bug with view rotation.

### LINKED ISSUE #156: Camera rotation does not work
`ipyvolume.pylab.view` method doesn't work, since `ipyvolume` uses threejs camera now.

OlehKSS:


maartenbreddels:
Awesome, thanks, I assume it's working

PR #163, opened 2018-08-16
Docs: Link to VTK was missing an http

hamogu:
and thus is appears as "https://ipyvolume.readthedocs.io/en/latest/www.vtk.org" in the rendered documentation which obviously does not work.


maartenbreddels:
Thanks!

PR #164, opened 2018-08-20
The revert of PR #145. Fix of alpha calculation of mesh fragment.

OlehKSS:


maartenbreddels:
We shouldn't premultiply ourselves, that's sth a user can do, that would be more flexible.

PR #169, opened 2018-09-15
New  multi-volume example

kmader:
multivolume example using PET-CT medical image (might make more sense than a head and astronomy image) it could also be taken one step further to be PET+CT+Label but that is harder to view

image


maartenbreddels:
Great example, thanks!

Issue #81, opened 2017-10-27
RGB volumes


VolkerH:
Hi,

excuse me if this is a silly question but I've been going through the example notebooks and the documentation and I haven't been able to figure out how to render RGB volumes, e.g. a numpy array of shape [nx, ny, nz, 3] where the last dimensions are RGB values.

If I interpret the examples correctly, most examples seem to plot single-valued volume data and apply some sort of intensity-based RGB lookup table.

maartenbreddels:
Indeed that is usually the case, still, interesting question.
However, do you mean [nx, ny, nz, 3] or [nx, ny, nz, 4], since without an opacity it would always render an opaque cube.

VolkerH:
Actually the data I have is microscopy fluorescence intensity data in multiple channels. I usually want to visualize up to three channels of them simultaneously, typically assigning one of the R,G,B channels to each of them.
The arrays I work with  therefore don't have an explicit opacity channel. That would have to be computed, maybe as the inverse of the total intensity in all channels. The (voxels that don't have fluorescence emission are transparent).

GenevieveBuckley:
Hi @maartenbreddels, I have the same use case as Volker so I'm going to jump in here. 

The kind of microscopy image data we'd be working with would generally arrive as numpy arrays with shape (plane, row, column, channel). Plane, row, and column are the three spatial dimensions (generally called z, y, and x). Channel is represented by color in the image volume. Note that you can have lots of channels - we're not just restricted to 3 for red, green and blue. Alpha is never specified in the data itself, that's something that's part of the particular transfer function you'd apply.

The multivolume rendering https://github.com/maartenbreddels/ipyvolume/pull/143 should make it totally feasible to display this kind of data. We could slice on the channel axis, and plot separate overlapping volumes with a different transfer function for each.

I haven't been able to get the branch with multivolume rendering working yet but if you can help me fix that, I can work on a pull request that does what I've described here.

maartenbreddels:
I think this can now be done using multivolume rendering. Render each channel as a separate volume, and give them 'ramp' transfer function, having a constant color (say red for the first, green for the second), and opacity increasing from 0 to 1.

GenevieveBuckley:
With a bit of fiddling I made this:
![ipyvolume_multivolshow](https://user-images.githubusercontent.com/30920819/46017417-c26c7b80-c11a-11e8-9dba-7f81933bab42.gif)

[...]

PR #178, opened 2018-09-23
[WIP] FEATURE: Convenience functions for volume rendering transfer functions

GenevieveBuckley:
I propose adding convenience functions for volume rendering transfer functions. 

- [x] single color linear ramp transfer functions
- [x] matplotlib colormaps as transfer functions
- [x] function for easy loading of a number of predefined transfer functions
- [ ] function tests - still in progress

Now that it's possible to have multiple volumes displayed in the same figure, we need something like this to increase the useability (and it looks like this has been suggested before):

> I think this can now be done using multivolume rendering. Render each channel as a separate volume, and give them 'ramp' transfer function, having a constant color (say red for the first, green for the second), and opacity increasing from 0 to 1.
https://github.com/maartenbreddels/ipyvolume/issues/81#issuecomment-420551030

I've also seen this issue play out too, since the default choice is very non-standard for my field:
> When I've explained the current system with the 3 RGB gaussians to my
professors, it's always taken a minute for them to understand that red is
low, green is medium, and blue is high. I think having a single linear
color gradient from low to high could eliminate that confusion.
https://github.com/maartenbreddels/ipyvolume/pull/76#issuecomment-336638057

Let me know what your thoughts are, happy to discuss etc.

### DIFF
--- a/ipyvolume/transferfunction.py
+++ b/ipyvolume/transferfunction.py
@@ -284,7 +284,7 @@ def load_transfer_functions(include_rgb_linear=True,
             rgb = colors[color_key]
             tf = linear_transfer_function(rgb)
             transfer_functions[color_key] = tf
-            reverse_tf = linear_transfer_function(rgb, reverse_opacity=True)
+            tf_reversed = linear_transfer_function(rgb, reverse_opacity=True)
             transfer_functions[color_key + '_r'] = tf_reversed
     # matplotlib colormaps
     if include_matplotlib:
--- a/ipyvolume/transferfunction.py
+++ b/ipyvolume/transferfunction.py
@@ -4,6 +4,7 @@ from traitlets import Unicode, validate
 from traittypes import Array
 import traitlets
 import numpy as np
+import matplotlib.cm
 from . import serialize
 from .serialize import array_rgba_png_serialization, array_serialization
 N = 1024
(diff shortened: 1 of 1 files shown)


GenevieveBuckley:
Travis CI build is currently failing due to the issues with unpkg (https://github.com/unpkg/unpkg.com/issues/134)

ipyvolume/transferfunction.py
```diff
@@ -164,3 +165,141 @@ def control(self, max_opacity=0.2):
...
+    if reverse_colormap: rgba = np.flip(rgba, axis=0)
+    # Create opacity values to overwrite default matplotlib opacity=1.0
+    opacity = np.linspace(min_opacity, max_opacity, num=_num_elements)
+    if reverse_opacity: opacity = np.flip(opacity, axis=0)
+    rgba[:,-1] = opacity # replace opacity=1 with actual opacity
+    transfer_function = TransferFunction(rgba=rgba)
+    return transfer_function
+
+
+def load_transfer_functions(include_rgb_linear=True,
```
maartenbreddels: I would not call it load (get?), and I'd make it 'singleton' like, so it will always return the same widgets, what do you think? (it will basically cache the widgets).
My guess is this will be used as part of a gui to select a particular colormap?

ipyvolume/transferfunction.py
```diff
@@ -164,3 +165,141 @@ def control(self, max_opacity=0.2):
...
+    >>> import ipyvolume as ipv
+    >>> rgb = (0, 255, 0)  # RGB value for green
+    >>> green_tf = ipv.transfer_function.matplotlib_transfer_function('viridis')
+    >>> ds = ipv.datasets.aquariusA2.fetch()
+    >>> ipv.volshow(ds.data[::4,::4,::4], tf=green_tf)
+    >>> ipv.show()
+
+    .. seealso:: linear_transfer_function()
+    """
+    _num_elements = 256  # length of rgba transfer function array
```
maartenbreddels: same here.

ipyvolume/transferfunction.py
```diff
@@ -164,3 +165,141 @@ def control(self, max_opacity=0.2):
 		return ipywidgets.VBox(
 			[ipywidgets.HBox([ipywidgets.Label(value="levels:"), l1, l2, l3]), ipywidgets.HBox([ipywidgets.Label(value="opacities:"), o1, o2, o3])]
 		)
+
+
+def linear_transfer_function(rgb_values,
```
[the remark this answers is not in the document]
GenevieveBuckley: That's much more elegant, thanks. Also means we can probably use this in conjunction with the color picker ipywidget, too! (Let me check that last part, but that would be great).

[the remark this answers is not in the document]
maartenbreddels: It would be great if we do it on the frontend, and indeed use the color picker!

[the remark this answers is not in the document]
maartenbreddels: What I've done in a previous project, was to put all the colormaps of matplotlib in 1 texture, of #texture x 1024 pixels, and then the index selected the row of the texture.

maartenbreddels:
I should also mention @vidartf 's work on ipyscales:
https://ipyscales.readthedocs.io/en/latest/examples/colorbar.html
would be really nice to integrate with that.

GenevieveBuckley:
Thanks for the link, I didn't know that about @vidartf. So far I've thought exactly zero about how to display colorbars next to figures, but that will be a pretty necessary next step.

ipyvolume/transferfunction.py
GenevieveBuckley: Let's call it 'predefined_transfer_functions'.

I'm not sure I understand correctly: do you want this to stay a function but always return the same dictionary, or have me make a singleton class for it?

I have no major plans for a colormap gui. But I do want to set future efforts at a GUI up for success as much as possible, so if you have advice on things I'm doing (or not doing here I'm happy to hear it.

What I want in the short term is a bunch of transfer function widgets so I can easily switch between them programmatically for the multivolume rendering I'm working on now.

maartenbreddels: What I would like to see, is when a transfer function is created with the same set of parameters (same key for the dictionary), that it returns the same transfer function. That will limit the number of widgets created, what do you think?

GenevieveBuckley: I'm not sure how best to implement the idea, but I think it's a good strategy/goal.

maartenbreddels: sth like the:
```
tf_cache = {}
def get_transfer_functions(...)
  tf = tf_cache.get(key)
  if not tf:
    tf = ....
    tf_cache[key] = tf
  ...
```

GenevieveBuckley:
@maartenbreddels can you explain the difference between the four different transfer function classes in python?

Those python classes are:
1. TransferFunction
2. TransferFunctionJsBumps
3. TransferFunctionWidgetJs3
4. TransferFunctionWidget3

Found in `ipyvolume/ipyvolume/transferfunctions.py`


And on the javascript side, there's
a. TransferFunctionModel
b. TransferFunctionJsBumpsModel
c. TransferFunctionWidgetJs3Model
d. TransferFunctionView

Found in `ipyvolume/js/tf.js`

PR #180, opened 2018-09-25
PEP8 imports

GenevieveBuckley:
I've done some tidying to bring the code more into line with PEP8, in particular the recommendations on imports.

## What this PR includes:
- [x] ordered imports according to the PEP8 guidelines (i.e. standard library imports first, third-party libraries second, ipyvolume imports last)
- [x] replaced wildcard imports with explicit imports 
- [x] moved all import statements to the top of the file

I haven't touched very much of examples.py or test_all.py, not sure if they need slightly special handling.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,8 @@ before_install:
   - conda info -a
   - conda create -q -n test-environment python=$PYTHON_VERSION numpy scipy runipy
   - source activate test-environment
-  - conda install -c conda-forge pytest pytest-cov bokeh
+  - conda install -c conda-forge pytest pytest-cov bokeh matplotlib scikit-image shapely
+  - pip install PyChromeDevTools
   - pip install coveralls
 
 install:
--- a/ipyvolume/__init__.py
+++ b/ipyvolume/__init__.py
@@ -1,13 +1,57 @@
 from __future__ import absolute_import
-from ._version import __version__
 
+from ._version import __version__
 from . import styles
-from .widgets import *
-from .transferfunction import *
 from . import examples
 from . import datasets
 from . import embed
-from .pylab import *
+from .widgets import (Mesh,
+                      Scatter,
+                      Volume,
+                      Figure,
+                      quickquiver,
+                      quickscatter,
+                      quickvolshow)
+from .transferfunction import (TransferFunction,
+                               TransferFunctionJsBumps,
+                               TransferFunctionWidgetJs3,
+                               TransferFunctionWidget3)
+from .pylab import (current,
+                    clear,
+                    controls_light,
+                    figure,
+                    gcf,
+                    xlim,
+                    ylim,
+                    zlim,
+                    xyzlim,
+                    squarelim,
+                    plot_trisurf,
+                    plot_surface,
+                    plot_wireframe,
+                    plot_mesh,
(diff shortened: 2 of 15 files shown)


GenevieveBuckley:
### The question of dependencies
Moving import statements to the top of each file has surfaced a few (possibly optional) dependencies. Should these be hard or optional dependencies?
* bokeh
* scipy
* scikit-image
* ChromeDevTools (this one looks like a debugging tool left in there)

If optional, does that mean those imports should be wrapped in a try/except, or handled in some other way?

GenevieveBuckley:
### Resolving inconsistencies
It's brought to light some of the inconsistencies that have confused me (both across ipyvolume as a whole, but there are also inconsistencies within the same file). Eg:

#### 1)
```python
import ipyvolume
import ipyvolume as ipv
```
I'd guess  `import ipyvolume as ipv` to be standard across the library, since this's what's used in all the tutorial examples I've seen. What do other people think here?

#### 2)
```python
import ipywidgets
import ipywidgets as widgets
```
I'd guess `import ipywidgets` should be preferred, since otherwise it could cause confusion with the ipyvolume widgets module. Thoughts?

#### 3)
```python
import ipyvolume.pylab as p3
from ipyvolume import pylab
```
I'd guess `from ipyvolume import pylab` should be preferred, since `p3` is pretty ambiguous. Is that right?

maartenbreddels:
> t's brought to light some of the inconsistencies that have confused me (both across ipyvolume as a whole, but there are also inconsistencies within the same file). Eg:
> 
> #### 1)
> ```python
> import ipyvolume
> import ipyvolume as ipv
> ```

The last one should be standard, p3 was what I used a long time ago.

maartenbreddels:
> Moving import statements to the top of each file has surfaced a few (possibly optional) dependencies. Should these be hard or optional dependencies?
> 
> * bokeh
> * scipy
> * scikit-image
> * ChromeDevTools (this one looks like a debugging tool left in there)
> 
> If optional, does that mean those imports should be wrapped in a try/except, or handled in some other way?

I prefer imports not at the top when they are optional.

maartenbreddels:
Nice work! Do you want me to merge this as it is?

GenevieveBuckley:
Sure go ahead and merge if you can confirm that these changes aren't the reason the TravisCI build is currently failing, 

Travis can't pip install ipyvolume, which is potentially quite serious. But it looks like it's something to do with webpack and the javascript, so I don't *think* my changes should have affected this?
```
The command "pip install -e ." failed and exited with 1 during .
```
```
Command "python setup.py egg_info" failed with error code 1 in /home/travis/build/maartenbreddels/ipyvolume/
```
... and in more detail:
```
> ipyvolume@0.5.2-dev.1 prepublish /home/travis/build/maartenbreddels/ipyvolume/js
    > webpack
    
    internal/crypto/hash.js:58
        throw new ERR_INVALID_ARG_TYPE('data',
        ^
    
    TypeError [ERR_INVALID_ARG_TYPE]: The "data" argument must be one of type string, TypedArray, or DataView. Received type undefined
```

maartenbreddels:
Yeah, that's bad, I have to look into it soon, many thanks for all the work you are doing!

PR #186, opened 2018-10-01
[WIP] FEATURE: Multivolume rendering for multi-dimensional data

GenevieveBuckley:
Extending volume rendering code to accomodate multi-dimensional data. Accepts either 3D or 4D data (3 spatial dimensions, plus color channels).

Example:
```python
import numpy as np
import ipyvolume as ipv

aquarius = ipv.datasets.aquariusA2.fetch() # 3D data
data = np.moveaxis(np.array([aquarius.data, aquarius.data.T]), 0, -1) # pseudo-4D data

ipv.quickvolshow(data)
```

ipyvolume_multivolshow

Unlike the existing brightness and opacity sliders, the colorpicker widget uses a python callback. This means the colorpicker widget won't work if there isn't a live python kernel (eg: if the ipyvolume widget is embedded in a html document). This is a disadvantage - is it possible to implement in javascript? I've seen [this previous discussion](https://github.com/jupyter-widgets/ipywidgets/issues/1109#issuecomment-299958315) but am unsure if there's been progress there.

Note: Any comments about the code generating the linear ramp transfer functions should go in the discussion at https://github.com/maartenbreddels/ipyvolume/pull/178

### DIFF
--- a/ipyvolume/pylab.py
+++ b/ipyvolume/pylab.py
@@ -21,7 +21,7 @@ try:
 except:
     from cStringIO import StringIO
 import base64
-from ipyvolume.transferfunction import linear_transfer_function as tf
+from ipyvolume.transferfunction import linear_transfer_function
 
 
 def _docsubst(f):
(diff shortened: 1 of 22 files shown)


ipyvolume/pylab.py
```diff
@@ -675,67 +674,71 @@ def volshow(data, lighting=False, data_min=None, data_max=None,
...
+                                                         description="opacity")
+            widget_brightness = ipywidgets.FloatLogSlider(base=10, min=-1, max=1,
+                                                         description="brightness")
+            widget_colorpicker = ipywidgets.ColorPicker(value=color,
+                layout=ipywidgets.Layout(width='15%'))
+            ipywidgets.jslink((vol, 'opacity_scale'), (widget_opacity_scale, 'value'))
+            ipywidgets.jslink((vol, 'brightness'), (widget_brightness, 'value'))
+            def change_transfer_function(vol, color):
+                vol.tf = linear_transfer_function(color.new)
+            widget_colorpicker.observe(lambda x, vol=vol: change_transfer_function(vol, x), names='value')
```
maartenbreddels: This would create a new widget for each color picked, I think this is a good reason to make this happen on the javascript side.

ipyvolume/pylab.py
```diff
@@ -675,67 +674,71 @@ def volshow(data, lighting=False, data_min=None, data_max=None,
     :param float data_min: minimum value to consider for data, if None, computed using np.nanmin
     :param float data_max: maximum value to consider for data, if None, computed using np.nanmax
     :parap int max_shape: maximum shape for the 3d cube, if larger, the data is reduced by skipping/slicing (data[::N]), set to None to disable.
-    :param tf: transfer function (or a default one)
+    :param tf_colornames: transfer function (or a default one)
```
maartenbreddels: What about letting tf be a string or tuple, like [in this PR](https://github.com/jupyter-widgets/ipywidgets/pull/2182/files) where the icon can be a string or Icon widget.

ipyvolume/pylab.py
```diff
@@ -675,67 +674,71 @@ def volshow(data, lighting=False, data_min=None, data_max=None,
...
+            def change_transfer_function(vol, color):
+                vol.tf = linear_transfer_function(color.new)
+            widget_colorpicker.observe(lambda x, vol=vol: change_transfer_function(vol, x), names='value')
+
+            widgets_bottom = [ipywidgets.HBox([widget_colorpicker, widget_opacity_scale, widget_brightness])]
+            current.container.children += tuple(widgets_bottom, )
+
+        fig.volumes = fig.volumes + [vol]
+
+    return fig
```
maartenbreddels: I'm against returning figure, since it does not follow conventions/intuition.

GenevieveBuckley: Return the full list of volumes, then? I'll do that.

ipyvolume/pylab.py
GenevieveBuckley: I'd like to do that. What's the javascript equivalent of observe?

maartenbreddels: I think it is best to look at this implementation:
https://github.com/maartenbreddels/ipyvolume/blob/9b7801b41bc3d77d96c01ef774854008b06b5353/ipyvolume/transferfunction.py#L79
https://github.com/maartenbreddels/ipyvolume/blob/9b7801b41bc3d77d96c01ef774854008b06b5353/js/src/tf.js#L57
And base it on that. Does that help?

ipyvolume/pylab.py
maartenbreddels: Well, check my above comment, why would volshow do this? Why not repeatedly call volshow?

maartenbreddels: Btw, for a quick chat, https://gitter.im/maartenbreddels/ipyvolume may be useful, and we could also do a videochat if that speeds things up.

GenevieveBuckley:
> What is the reason for doing this in volshow, why not simply a loop/multiple calls? 

Because it's a pain in the butt, that's why. I want to write a single funciton call, rather than a bunch of boilerplate *every time* I need to visualize an array. I'm generally working with multidimensional data in a 4D or 5D numpy array. That's also true for most other people working with microscopy images, hyperspectral astronomical data, medical imaging, etc. So there are a lot of use cases for extending this.

> I think it may make volshow's behaviour too complex, and maybe less flexible even.

What flexibility do you think is lost? You can still pass in a single 3D arrray and have it work the same as before. (In fairness, this PR mucks around with the default transfer function, but that's the only change, and even that might not stay like that for long)

evolu8:
Has this been abandoned?

GenevieveBuckley:
Hi @evolu8, this fell off my radar after the discussion about the different types of transfer function widgets in https://github.com/maartenbreddels/ipyvolume/pull/178. Did you want to take a crack at this?

PR #189, opened 2018-10-04
Fix doubled command in README

maxstrobel:
In the instructions of the jupyter-lab setup is a command doubled.


maartenbreddels:
Thanks Max!

maartenbreddels:
I don't think that error has anything to do with this PR btw

Issue #195, opened 2018-10-30
Select points from one of multiple scatter objects


jmeyers314:
Selection works great when there's only one widget(?) being plotted:

```python
import ipyvolume as ipv
import numpy as np

ipv.figure()
s=ipv.scatter(
    np.random.uniform(size=10),
    np.random.uniform(size=10),
    np.random.uniform(size=10),
    marker='sphere', size=1, color='red'
)
ipv.selector_default()
ipv.show()
```

but when I try putting in a second widget like so:

```python
import ipyvolume as ipv
import numpy as np

ipv.figure()
s=ipv.scatter(
    np.random.uniform(size=10),
    np.random.uniform(size=10),
    np.random.uniform(size=10),
    marker='sphere', size=1, color='red'
)
s2=ipv.scatter(
    np.linspace(0, 1., 10),
    np.linspace(0, 1., 10),
    np.linspace(0, 1., 10),
    marker='sphere', size=1, color='blue'    
)
ipv.selector_default()
ipv.show()
```

trying to select a point or points from either widget yields:

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
~/src/lsstsw/miniconda/envs/lsst-scipipe/lib/python3.6/site-packages/ipyvolume/pylab.py in lasso(data, other, fig)
   1187                 x, y = fig.project(scatter.x, scatter.y, scatter.z)
   1188                 mask = inside(x, y)
-> 1189                 scatter.selected = join(scatter.selected, np.where(mask), fig.selection_mode)
   1190     fig.on_selection(lasso)
   1191 

~/src/lsstsw/miniconda/envs/lsst-scipipe/lib/python3.6/site-packages/ipyvolume/pylab.py in join(x, y, mode)
   1166                     return (x > xmin) & (x < xmax) & (y > ymin) & (y < ymax)
   1167             def join(x, y, mode):
-> 1168                 N = np.max(x) if x is not None else np.max(y)
   1169                 N = max(N, np.max(y))
   1170                 xmask = np.zeros(N+1, np.bool)

~/src/lsstsw/miniconda/envs/lsst-scipipe/lib/python3.6/site-packages/numpy/core/fromnumeric.py in amax(a, axis, out, keepdims)
   2270 
   2271     return _methods._amax(a, axis=axis,
-> 2272                           out=out, **kwargs)
   2273 
   2274 

[...]

PR #197, opened 2018-10-30
Fix bug on empty point selection

jmeyers314:
Can't use np.max() on empty array, so handle that case explicitly.

Addresses #195


jmeyers314:
Oops.  Looks like travis failed to download something.  Restart please?

maartenbreddels:
Looks good enough for me! :)

maartenbreddels:
Many thanks, and thanks for the first time contribution!

Issue #192, opened 2018-10-21
offline=True in embed_html failing to save required file js/jupyter-threejs.js


GenevieveBuckley:
The `offline=True` kwarg in `ipyvolume.embed.embed_html()` is failing to save the file at https://unpkg.com/jupyter-threejs@2.0.1/dist/index.js locally as `js/jupyter-threejs.js`

Copying that file manually into `js/jupyter-threejs.js` fixes the issue and allows offline viewing.

I'm using ipyvolume installed from master branch: 0.5.2-dev.1

GenevieveBuckley:
It's possible this is wholly or partially resolved by the offline_embed branch here: https://github.com/maartenbreddels/ipyvolume/tree/offline_embed

maartenbreddels:
That branch is old, I'll remove it. Seems like a legit bug.

GenevieveBuckley:
Maybe related: I was wondering if it would be of general use to have a no dependencies option for the offline html? It'd be useful in situations where you write many html files to the same location, and care about how fast you can write them to disk. 
(This is something we're doing at the moment, overwriting a single file with new data, and I think it'd speed up the responsiveness quite a bit if it didn't re-download every dependency each time.)

maartenbreddels:
Yes, a cache would be nice, I think though it should (optionally) check the CDN for the right version, and if not it should update it. My guess is that the request package can do this. I also hate that it downloads everything every time. It can also do a hardlink to the cached file (for non-windows users).

PR #198, opened 2018-11-01
BUGFIX: offline=True kwarg in embed_html was failing to save dependency jupyter-threejs.js

GenevieveBuckley:
This PR adds a function `save_jupyterthreejs` to `embed.py` to download and save the `jupyter-threejs.js` dependency required for offline viewing of ipyvolume html files.

Closes https://github.com/maartenbreddels/ipyvolume/issues/192

> The offline=True kwarg in ipyvolume.embed.embed_html() has been failing to save the file at https://unpkg.com/jupyter-threejs@2.0.1/dist/index.js locally as js/jupyter-threejs.js
Copying that file manually into js/jupyter-threejs.js fixes the issue and allows offline viewing.


ipyvolume/_version.py
```diff
@@ -3,4 +3,6 @@
 __version__ = '0.5.2-dev.1'
 __version_tuple_js__ = (0, 5, 2, 'dev.1')
 __version_js__ = '0.5.2-dev.1'
-__version_threejs__ = '0.91' # kept for embedding in offline mode, we don't care about the patch version since it should be compatible
\ No newline at end of file
+__version_threejs__ = '2.0.1'
```
maartenbreddels: threejs is at version 98 btw: https://threejs.org/
pythreejs is what you meant, I know, it is confusing.
I'll talk with @vidartf about it, how we can get the right threejs version that matches the pythreejs one.

maartenbreddels:
As it is now, threejs is embeded in jupyter-threejs's embed.js, although it shouldn't, and it is actually configured using require's path in https://github.com/jupyter-widgets/pythreejs/blob/master/js/src/embed.js
@vidartf 's plan is to test making threejs external, exposing the semver version in the Python package, and then we know if it works.
Also, ipyvolume should do the same, make the threejs external, and copy the config of https://github.com/jupyter-widgets/pythreejs/blob/master/js/src/embed.js 

@GenevieveBuckley it might be that the PR as it is now is working, can you confirm that? ipyvolume and jupyter-threejs have their own version of threejs, which is non-ideal, but if it works, it works.

PR #228, opened 2019-02-11
Faster js builds + karma

maartenbreddels:
The build was setup to use ts-loader, while the sourcecode was already in the lib directory. Now build times are much faster.
Karma setup is more clear and includes source mapping, although it's not fully working yet for the files going throught ts-loader (the error msges point to the right .ts file, but the line number of the .js file).


wolfv:
gr8!

maartenbreddels:
50-100x speedup if you use these:
```
(cd js; npm run watch) # terminal 1
(cd js; ./node_modules/.bin/karma start --no-single-run --browsers Chrome) # terminal 2
```
Edit src/*.js or test/*.ts and an instant rebuild, an near instant execution of test suite.

wolfv:
for me it's only really fast if I don't use watch, but only `watch:nbextension` otherwise the typescript compilation eats quite some time (1800ms vs 80ms) but good stuff!

maartenbreddels:
That's odd, but you do need to run the full watch command, otherwise you can only edit the lib directory (which is generated).. Here tsc in watch mode is near lightspeed, I don't even see it taking time.

wolfv:
ready to merge?

PR #248, opened 2019-05-01
Include license in wheels

toddrme2178:
The MIT license requires all copies of the software include the license.


maartenbreddels:
Thanks a lot for your first contribution here!

Issue #89, opened 2017-12-14
OrbitControl behaves strangely after first drag


mathieu1:
Using `fig.camera_control = 'orbit'`, the Orbital Control works as expected at first glance. However, after dragging the object and "tilting the horizontal plane" to a "non-horizontal position", the orbit controls do not behave as expected anymore, and do not keep the Y axis upwards (contrary to the standard behaviour of threejs' OrbitControl)

Hopefully what I mean is clarified with this:
![45zkr-ev7z3](https://user-images.githubusercontent.com/7703352/34018288-7cb0a714-e129-11e7-9965-e5dd869e68c0.gif)

maartenbreddels:
Thanks, and this is only using *that* camera control right? Not switching between them?

mathieu1:
Indeed, nothing else happens in that clip besides two consecutive mouse drags. The code has nothing special AFAICT (nevermind the x,y,z rearrangement):
```python
fig = p3.figure()
fig.camera_control = 'orbit' 
p3.plot_trisurf(y, z, x, triangles=faces, color=colors)
p3.show()
```

Carreau:
Same for me, regardless of whether `orbit` is set from the python side, or by clicking the "up arrow" button. (BTW, I would suggest changing it for "globe") that would be IMHO more representative.

mgeier:
@mathieu1 and @Carreau Does it work better with #273?

PR #273, opened 2019-09-24
Enable panning in "orbit" mode

mgeier:
To select "orbit" mode, click the "up arrow" symbol in the widget.

Panning works with the right mouse button.

Panning with the arrow keys on the keyboard is still disabled (which I think was the reason to disable panning in the first place).

I had to remove some `lookAt()` calls, but I don't see any change in the behavior of the "trackball" mode because of this. Did I miss something?

This is a (less intrusive) subset of #272.

### DIFF
--- a/js/src/figure.js
+++ b/js/src/figure.js
@@ -395,13 +395,6 @@ var FigureView = widgets.DOMWidgetView.extend({
         if (this.model.get('camera')) {
             this.camera = this.model.get('camera').obj
             this.model.get('camera').on('change', () => {
-                // the threejs' lookAt ignore the quaternion, and uses the up vector
-                // we manually set it ourselve
-                var up = new THREE.Vector3(0, 1, 0);
-                up.applyQuaternion(this.camera.quaternion);
-                this.camera.up = up;
-                this.camera.lookAt(0, 0, 0);
-                // TODO: shouldn't we do the same with the orbit control?
                 this.control_trackball.position0 = this.camera.position.clone();
                 this.control_trackball.up0 = this.camera.up.clone();
                 // TODO: if we implement figure.look_at, we should update control's target as well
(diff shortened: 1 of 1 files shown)


mgeier:
I think this might even fix #89.

js/src/figure.js
```diff
@@ -634,7 +627,10 @@ var FigureView = widgets.DOMWidgetView.extend({
         this.control_orbit = new THREE.OrbitControls(this.camera, this.renderer.domElement);
         this.control_trackball.dynamicDampingFactor = 1.
         this.control_trackball.noPan = true;
-        this.control_orbit.enablePan = false;
+        // All widgets on the page would receive those keypresses all the time:
```
maartenbreddels: I am not sure what you mean by this comment.

Tronic:
I recently rewrote my vis code straight for pythreejs largely because of camera handling and found out that OrbitControls just work perfectly, even for panning, with no tweaks required. Hopefully this helps.

js/src/figure.js
mgeier: I just wanted to explain why `enableKeys` is switched off:

The problem is that if there are multiple `ipyvolume` widgets on the same page and I press, e.g., the left arrow key, all of them pan to the left. Which is most likely not the expected behavior.

But I can simply remove the comment if it causes confusion.

... or replace it by whatever you prefer. Just tell me.

mgeier:
@Tronic I was also using `pythreejs` in the meantime and made the same observation.

I guess when `ipyvolume` was developed, `pythreejs` didn't yet have those features?

@maartenbreddels Are there plans to re-write some of `ipyvolume` using those new `pythreejs` features?

maartenbreddels:
https://github.com/maartenbreddels/ipyvolume/pull/280 replaces this PR

mgeier:
If I'm understanding this correctly, #280 isn't actually a replacement for this PR, see https://github.com/maartenbreddels/ipyvolume/pull/280#issuecomment-552415725.

PR #280, opened 2019-10-29
Feature: pythreejs controls

maartenbreddels:
@mgeier: I gave you credits for the 2nd commits.

Replaces #273 by allowing a pythreejs Controls. Example usage:

```python
import pythreejs
import ipyvolume as ipv
import numpy as np
fig = ipv.figure()
ipv.plot_trisurf([0, 0, 3., 3.], [0, 4., 0, 4.], 2,
           triangles=[[0, 2, 3], [0, 3, 1]])
ipv.show()

control = pythreejs.OrbitControls(controlling=fig.camera)
fig.controls = control
control.autoRotate = True
# the controls does not update itself, but if we toggle this setting, ipyvolume will update the controls
# fig.render_continuous = True
```

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -136,6 +136,7 @@ Contents
     animation
     api
     vr
+    pythreejs
 
 
 Changelog
--- a/docs/source/pythreejs.ipynb
+++ b/docs/source/pythreejs.ipynb
@@ -1,0 +1,155 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Integration with pythreejs\n",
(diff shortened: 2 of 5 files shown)


mgeier:
Do I understand this correctly:

This allows users to explicitly specify (properly functioning) `pythreejs.OrbitControls`, but it doesn't change anything in the default behavior?

If yes, this is not at all what I intended to do in #273.
I wanted the default to be reasonable.
And in the long run, I wanted to try to make "orbit" controls the default, because I don't like this "trackball" thing.

Being able to explicitly specify `OrbitControls` is a great feature, but something quite different.

maartenbreddels:
> This allows users to explicitly specify (properly functioning) `pythreejs.OrbitControls`, but it doesn't change anything in the default behavior?

yes. 

Commented noted, I think this gives users more flexibility, without changing the current behavior, which I think is fine.

PR #305, opened 2020-04-01
Feature: 3d bar charts

maartenbreddels:
This enables tweaking the scatter object a bit, so that we can make bar charts, two examples are included, a normal bar chart:
bars

Or on a sphere:

ipyvolume-spherical-barchart

It would be nice to combine the spherical bar chart with some geojson data.


JovanVeljanoski:
That is incredible!! Great idea!

maartenbreddels:
Thanks, a slightly prettier version:
ipyvolume-bars-spherical

Will update and rebase the PR later.

Issue #276, opened 2019-10-08
pythreejs Low Level Modifications (e.g. add threeJS Spot_Light)


TimoFriedri:
m using your library extensively for the plotting of 3D objects (meshes and voxel models).

Unfortunately, the lightning is quite simple and it is hard to infer the shape from some angles. 
I'd like to add additional light sources and activate shadows, which seems to be possible from the pythreejs side.

So far I have not been successful is doing so. It is quite hard to figure out what I have to edit/add ... in the respective scenes/meshes/ ... which are stored in a figure.widgets container.

Could someone give some hints on how to achieve "adding and modifying lightning for ipyvolume plot?

Maybe an low level modification guide in the docs would be nice.

Thanks

maartenbreddels:
Hi Timo,

I think we have to change the shaders to take into account the pythreejs lighting. That involves diving into threejs, I am happy to assist you with the ipyvolume side of things.
Basically `js\glsl\*`needs to be  edited, and Figure should probably be an Scene subclass, or Object3D, so it can have lights, and then follow something along these lines:
https://stackoverflow.com/questions/30151086/threejs-how-do-i-make-a-custom-shader-be-lit-by-the-scenes-lights.
Do you feel like attacking this issue?

cheers,

Maarten

rinftech-github:
Greetings!

We are working on adding Lighting(Ambiental, Directional, Spot, Point, Hemisphere) support for ipyvolume rendering.
Below we are describing how we are planning to go about it.

Similar to pylab functions like plot_mesh, we are adding functions for every light type(exp addDirectionalLight()), each with specific parameters (some being adjsutable through sliders and color pickers).
Further, a Light widget class is necessary with associated TypeScript classes(LightView, LightModel).

Our current shader tests feature vertex and fragment code from three.js(Lambert, Phong, Physical), combined with existing ipyvolume shader code, in order to be up to the visual standard from three.js examples. 

What is your opinion regarding the plan above?
Are there any shader modules that should not be present in the new code? (exp: skinning, morphtargets)
We are planning to specify as parameter the target scene in which a specific light will be added in order to diferentiate between scene_scatter, scene_volume, scene_opaque. Do you agree with this approach?

Please, if you have time, can you give any feedback in order for us to better prepare for the eventual Pull Request?

rinftech-github:
We are concentrating on this until the end of March 2020. So we would appreciate any feedback as soon as possible. Thank you!

[...]

PR #308, opened 2020-04-06
Feature: Lights and Shadows

TimoFriedri:
Hi,

we have added light and shadow support to ipyvolume.
We tried to not interfere with the existing functionality of ipyvolume,
As long as no lights are added, nothing changes for the user.
As soon as a light is added to the Figure (Scene), we change to our custom shaders. But this can also be reversed on the fly by just setting the lighting_mode.

The status is as follows:
**plot_mesh, plot_trisurf, plot_surface**
These 3 fully support light and shadows for:
spot_light, point_light, directional_light, hemisphere_light and ambient_light

**scatter**
We only have a working implementation for hemisphere and ambient_light.
Due to a bug in the old threejs version, this does not work correctly.
Please have a look at the issue for more information.
#276 

We have added a Notebook with examples for the usage of the new functionality.
Unittests are added.
Inline doc should be present.


maartenbreddels:
Amazing!, give me some time to digest/review this, a first question after taking a quick look:

Have you considered using https://pythreejs.readthedocs.io/en/stable/api/lights/index.html instead of our own Light widget?

TimoFriedri:
Uff, maybe @rinftech-github  can give some more details about that.

Something that comes into my mind is the automatic shader switch as soon as a light is added.
Scatter and plot_trisurf ... have now functions to switch the shader after they have been added to the figure on the javascript side. Not sure how this would be done with the pythreejs wrapper. 
... *thinking* ...
hmm, probably would also be possible to update the plots on the python side as this (lighting_model) is just another traitlet class value.

rinftech-github:
Hello
Thank you for your feedback!
The main reason for not adding lights on python side is because I wasn't able to access the scatter scene.

TimoFriedri:
The inaccessible scene was the reason I started issue #276 because my I couldn't gt pythreejs lights to work.

js/glsl/mesh-fragment-lambert.glsl
```diff
@@ -0,0 +1,124 @@
...
+    uniform sampler2D texture_previous;
+    uniform float animation_time_texture;
+#endif
+
+void main() 
+{
+	vec4 finalColor2 = vec4( 0.0, 0.0, 0.0, 1.0 );
+	
+#ifdef USE_RGB
+    finalColor2 = vec4( vertex_color.rgb, 1.0 );
```
maartenbreddels, line 54: I think this color will be used for the lighting right? in the RGB pass it is important that we just render the vertex_color which represents the normalized xyz coordinates. These coordinates are used in the volume rendering pass as 'start' and 'end' coordinates.

maartenbreddels: Also, I guess the finalColor2 is used in the threejs shader, do you have a pointer to some docs about this?

js/glsl/mesh-vertex.glsl
```diff
@@ -23,11 +23,9 @@ attribute vec3 position_previous;
     varying vec2 vertex_uv;
 #endif
 
-attribute vec4 color;
+attribute vec4 color_current;
```
maartenbreddels, line 26: Was there a need to rename this? Does it collide with threejs' naming?

js/src/mesh.ts
```diff
@@ -54,13 +78,31 @@ class MeshView extends widgets.WidgetView {
...
+        ] );
+
         const get_material = (name)  => {
             if (this.model.get(name)) {
                 return this.model.get(name).obj.clone();
             } else {
                 const mat = new THREE.ShaderMaterial();
                 mat.side = THREE.DoubleSide;
+
+                mat.flatShading = false;
```
maartenbreddels, line 102: Why is this needed?

js/src/mesh.ts
```diff
@@ -54,13 +78,31 @@ class MeshView extends widgets.WidgetView {
...
+                mat.transparent = true;
```
maartenbreddels, line 103: Why do we need to set tranparant to True? Can't we use Mesh.material.transparent ?

js/src/mesh.ts
```diff
@@ -90,11 +132,21 @@ class MeshView extends widgets.WidgetView {
...
         this.model.on("change:texture", this._load_textures, this);
         this.model.on("change:visible", this.update_visibility, this);
+        this.model.on("change:lighting_model change:opacity change:specular_color change:shininess change:emissive_color change:emissive_intensity change:roughness change:metalness change:cast_shadow change:receive_shadow", 
+        this.update_visibility, this);
+    }
+
+    public force_lighting_model() {
+        if(this.lighting_model === this.LIGHTING_MODELS.DEFAULT){
+            this.model.set("lighting_model", this.LIGHTING_MODELS.PHYSICAL);
+            this.update_visibility();
```
maartenbreddels, line 148:
```suggestion
            this. _update_materials();
```
I guess that should be enough right?

js/src/mesh.ts
```diff
@@ -266,23 +332,67 @@ class MeshView extends widgets.WidgetView {
...
+                material.fragmentShader = require("raw-loader!../glsl/mesh-fragment-lambert.glsl");
+            }
+            else if(this.lighting_model === this.LIGHTING_MODELS.PHONG) {
+                material.vertexShader = require("raw-loader!../glsl/mesh-vertex-phong.glsl");
+                material.fragmentShader = require("raw-loader!../glsl/mesh-fragment-phong.glsl");
+            }
+            else if(this.lighting_model === this.LIGHTING_MODELS.PHYSICAL) {
+                material.vertexShader = require("raw-loader!../glsl/mesh-vertex-physical.glsl");
+                material.fragmentShader = require("raw-loader!../glsl/mesh-fragment-physical.glsl");
+            }
```
maartenbreddels, line 367: I think that instead of having 4 shader, we can put them in 1, and use the pre-processor + defines to control the specifics.

js/src/mesh.ts
```diff
@@ -266,23 +332,67 @@ class MeshView extends widgets.WidgetView {
...
             material.depthWrite = true;
             material.transparant = true;
+            material.transparent = true;
```
maartenbreddels, line 370:
```suggestion
```

maartenbreddels: duplicate.

js/src/mesh.ts
```diff
@@ -266,23 +332,67 @@ class MeshView extends widgets.WidgetView {
...
+            else if(this.lighting_model === this.LIGHTING_MODELS.PHONG) {
+                material.vertexShader = require("raw-loader!../glsl/mesh-vertex-phong.glsl");
+                material.fragmentShader = require("raw-loader!../glsl/mesh-fragment-phong.glsl");
+            }
+            else if(this.lighting_model === this.LIGHTING_MODELS.PHYSICAL) {
+                material.vertexShader = require("raw-loader!../glsl/mesh-vertex-physical.glsl");
+                material.fragmentShader = require("raw-loader!../glsl/mesh-fragment-physical.glsl");
+            }
             material.depthWrite = true;
             material.transparant = true;
```
maartenbreddels, line 369: I wonder why *I* set this to true by default actually.

js/src/scatter.ts
```diff
@@ -163,6 +202,14 @@ class ScatterView extends widgets.WidgetView {
             );
         }
     }
+
+    public force_lighting_model() {
+        if(this.lighting_model === this.LIGHTING_MODELS.DEFAULT){
+            this.model.set("lighting_model", this.LIGHTING_MODELS.PHYSICAL);
+            this.update_visibility();
```
maartenbreddels, line 209:
```suggestion
            this. _update_materials();
```
similar to mesh?

js/src/scatter.ts
```diff
@@ -319,11 +415,12 @@ class ScatterView extends widgets.WidgetView {
         }
         const sprite = geo.endsWith("2d");
         const buffer_geo = new THREE.BufferGeometry().fromGeometry(this.geos[geo]);
+        buffer_geo.computeVertexNormals();
```
maartenbreddels, line 418: Do we always need this? Does this impact performance? Can we instead move this to the constructor?

js/src/scatter.ts
```diff
@@ -319,11 +415,12 @@ class ScatterView extends widgets.WidgetView {
...
         const instanced_geo = new THREE.InstancedBufferGeometry();
 
         const vertices = (buffer_geo.attributes.position as any).clone();
         instanced_geo.addAttribute("position", vertices);
-
+
```
maartenbreddels, line 423:
```suggestion
```

js/src/scatter.ts
```diff
@@ -335,6 +432,10 @@ class ScatterView extends widgets.WidgetView {
         const current  = new values.Values(scalar_names, [], this.get_current.bind(this), sequence_index, vector4_names);
         const previous = new values.Values(scalar_names, [], this.get_previous.bind(this), sequence_index_previous, vector4_names);
 
+        // Workaround for shader issue - color redefine 
+        current.ensure_array(["color"]);
```
maartenbreddels, line 436: I think that anwsers a previous comment, maybe also put it in the shader code, sth like `// threejs already uses the name color `

js/src/scatter.ts
```diff
@@ -335,6 +432,10 @@ class ScatterView extends widgets.WidgetView {
...
+        instanced_geo.addAttribute("color_current", new THREE.BufferAttribute(current.array_vec4.color, 4));
```
maartenbreddels, line 437: I don't fully understand the solution here, I'm afraid it will upload the color array twice to the GPU now, as 'color' and 'color_current'. I wonder if we should rename all `x` and `x_previous` to `x_next` and `x_previous` to avoid name collisions.

ipyvolume/widgets.py
```diff
@@ -217,7 +238,50 @@ def _update_data(self):
         self.data = np.array(data_view)
         self.extent = extent
 
+@widgets.register
+class Light(widgets.Widget):
```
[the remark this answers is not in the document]
rinftech-github: Hello. The main issue here is that the scatter scene can't be accessed from python in order to add lights. That is why I added a light widget class. 
Can the scatter scene be accessed from python? 
Thank you :)

js/glsl/mesh-fragment-lambert.glsl
rinftech-github: The shader code must be optimized/refactored.

js/glsl/mesh-vertex.glsl
rinftech-github: There already exists a variable named 'color' in the shader code from three.js. I don't remember exactly in which include, but yes, it was necessary. 
I will go into more depth when we being refactoring/fixing the code.

js/src/mesh.ts
rinftech-github: Sounds good. Do you think that support for Lambert, Phong and Phyisical lighting models is exhaustive?

js/src/mesh.ts
rinftech-github: Refactoring is necessary.

js/src/mesh.ts
TimoFriedri: I have recently seen dome eyed lighting. Maybe another one we can consider. It's e.g. in VTK/pyvista.

js/src/scatter.ts
rinftech-github: I believe the 'ensure_array' line is not longer necessary. I must study the scatter.ts code/behavior more before I give a definite answer.

js/src/scatter.ts
rinftech-github: Yes, I reckon this line can be moved to the constructor.

js/glsl/mesh-vertex.glsl
rinftech-github: **uniform vec3 color** is defined in shadow_frag.glsl and again in removed line **attribute vec4 color**

js/src/mesh.ts
rinftech-github: not needed. removing

js/src/mesh.ts
rinftech-github: fixing

js/src/scatter.ts
rinftech-github: updated code

js/src/mesh.ts
rinftech-github: merged into mesh-fragment and mesh-vertex

rinftech-github:
https://github.com/maartenbreddels/ipyvolume/pull/341
Integrated changes into branch feat_lighting_merge.
Removed light widget.
Encountered scaling issues not introduced by my code.
This PR can be closed.

PR #314, opened 2020-04-16
import Mapping from abc

glyg:
Importing and ABC directly from collections is deprecated and raises a warning in my tests.
This should fix the warning


maartenbreddels:
Thanks!

glyg:
No pb!
You saw that python2.7 breaks right? Not that it matters really, but maybe you can cut it off you travis?

maartenbreddels:
Ouch, that is really sloppy of me. Would you mind opening a PR for that? I wasn't planning to support it anymore.

PR #316, opened 2020-04-16
drop 2.7, add 3.8

glyg:
I'm using 3.8 already, IDK if you want to put it in the matrix

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,9 @@ os:
  - osx
 env:
  matrix:
-  - PYTHON_VERSION=2.7
   - PYTHON_VERSION=3.6
   - PYTHON_VERSION=3.7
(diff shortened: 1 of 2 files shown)


maartenbreddels:
Yes, good idea!

glyg:
btw I use pytest with nbval to run notebooks as tests (to replace runipy if needed)

glyg:
Ok it passed (except for a small lowering of the coverage).

Should stating that python 2.7 is not supported be made clear elsewhere?

maartenbreddels:
I'll do the changelog afterwards, so I'll see this in the gitlog!

I'm working on setting up testing with screenshots, that should replace the notebook testing.

maartenbreddels:
Thanks!

PR #354, opened 2020-10-07
Import ABC from collections.abc explicitly.

tirkarthi:
Directly accessing `abc` module from `collections` raises error. This worked since one of the other modules would have import `collections.abc`

```
python                                                       
Python 3.9.0b5 (default, Jul 20 2020, 20:46:56) 
[GCC 7.5.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import collections
>>> collections.abc.Mapping
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.9/collections/__init__.py", line 68, in __getattr__
    raise AttributeError(f'module {__name__!r} has no attribute {name!r}')
AttributeError: module 'collections' has no attribute 'abc'
```

### DIFF
--- a/ipyvolume/utils.py
+++ b/ipyvolume/utils.py
@@ -5,6 +5,7 @@ import io
 import time
 import functools
 import collections
+import collections.abc
 
 import numpy as np
 import requests


maartenbreddels:
Thanks!

PR #355, opened 2020-11-05
Support for JupyterLab 3.0 using Federated Bundles

maartenbreddels:
* Similar to https://github.com/bqplot/bqplot/pull/1222
 * setup.py could be reduce if https://github.com/jupyter/jupyter-packaging/pull/54 is released
 * Does not work yet if jupyter-widget is a Federated bundle, and ipydatawidgets and pythreejs are not.


maartenbreddels:
Thanks @martinRenou for the the example at https://github.com/bqplot/bqplot/pull/1311 and @mbektasbbg for making Galata!

PR #361, opened 2021-01-02
fix headless rendering with latest chrome

rafique:


ipyvolume/headless.py
```diff
@@ -49,7 +49,20 @@ def _screenshot_data(
...
     #     assert isinstance(fig, ipv.Figure)
 
-    chrome = PyChromeDevTools.ChromeInterface(headless_kwargs)
+
+    try:
+        host = "localhost"
+        port = 9222
+
+        if headless_kwargs.get("host"):
+            host = headless_kwargs.get("host")
```
maartenbreddels, line 58: i guess we can make host and port arguments of this function.

ipyvolume/headless.py
```diff
@@ -49,7 +49,20 @@ def _screenshot_data(
...
+
+        chrome = PyChromeDevTools.ChromeInterface(host=host, port=port)
+        #chrome = PyChromeDevTools.ChromeInterface(host="localhost")
```
maartenbreddels, line 61: please remove

ipyvolume/headless.py
```diff
@@ -59,16 +72,19 @@ def _screenshot_data(
     result = chrome.Runtime.evaluate(expression='ipvss()')
     tries = 0
     while tries < 10:
+        #print(str(result))
         try:
-            url = result['result']['result']['value']
+            url = result[0]['result']['result']['value']
```
maartenbreddels, line 77: why did this change, and which version, and can we make it detect this? so it also works on older versions?

ipyvolume/headless.py
```diff
@@ -59,16 +72,19 @@ def _screenshot_data(
...
             return url
-        except:
-            if 'ipvss' in result['result']['result']['description']:
+        except Exception as ex:
+            #print(str(ex))
+            #raise ex
```
maartenbreddels, line 81: please remove

ipyvolume/headless.py
```diff
@@ -59,16 +72,19 @@ def _screenshot_data(
...
+            if 'ipvss' in result[0]['result']['result']['description']:
```
maartenbreddels, line 82: maybe better to put result into a variable.

ipyvolume/headless.py
```diff
@@ -59,16 +72,19 @@ def _screenshot_data(
...
                 tries += 1
                 time.sleep(0.5)
             else:
                 print('error getting result, return value was:', result)
-                raise
+                raise ex
```
maartenbreddels, line 87: a bare raise should do the same thing right?

maartenbreddels (changes requested):
Hi,

thanks for this! I have a few comments, hope you can address them.

Regards,

Maarten

ipyvolume/headless.py
rafique: Actually I don't know why/when this has changed. This seems to work fine but recently throwing the following error: "tuple indices must be integers or slices, not str". This is the main reason prompting this pull request.

PR #372, opened 2021-03-22
Update dependencies for jupyterlab_3 branch

agoose77:
I needed to bump the webpack dependencies to be able to install via pip for JupyterLab 3.

I'm not an expert in the JS ecosystem, so let me know if anything here is superfluous / the updates are too aggressive.


maartenbreddels:
I think this can be closed now that https://github.com/maartenbreddels/ipyvolume/pull/355 is released.

maartenbreddels:
Thanks for the work!

agoose77:
Excellent news, thank you for fixing!

PR #388, opened 2021-09-07
rtd: jupyter_client issues

maartenbreddels:


maartenbreddels:
See https://github.com/jupyter/jupyter_client/issues/637

PR #418, opened 2022-11-09
Update header in lighting.ipynb

tmcaloone:
Was looking through the docs and noticed "Lighting" has been misspelled as "Lightning" in the page header.


maartenbreddels:
Thank you!

PR #424, opened 2023-01-27
Fix CI

maartenbreddels:


maartenbreddels:
docs failure will be fixed when the new ipyvolume is released

PR #442, opened 2023-11-29
feat: option to hide side panel

iisakkirotko:


astrofrog:
We need this over at glue-jupyter (https://github.com/glue-viz/glue-jupyter/issues/449), so +1 from me!

PR #448, opened 2025-04-30
fix: support numpy2 compatibility

flokli:
Since numpy2 we need to explicitly decide how to deal with values out of range.


maartenbreddels:
I think that's a reasonable fix. Is this compatible with numpy 1? If not, can we make it compatible?

flokli:
Both functions exists in 1.26, so I'd say yes. It's a bit hard to test for me in this environment, as numpy2 propagates from all other inputs as well and would need to be replaced.

flokli:
I tried to do the overrides in downstream dependencies (bqplot, bqscales, pandas), but it's [not possible to build pandas with numpy 1 anymore](https://github.com/pandas-dev/pandas/blob/v2.2.3/pyproject.toml#L11), making it really hard for me to test here.

Are you able to run the tests from your side, with a numpy 1 in the environment?

